PR #229, opened 2017-08-27
Added Label and sticky properties in Object definitions

svchandra:
Added stick and label.

### DIFF
--- a/developing-connectors/sdk/object-definition.md
+++ b/developing-connectors/sdk/object-definition.md
@@ -73,7 +73,7 @@ object_definitions: {
         </tr>
         <tr>
             <td>stick</td>
-            <td>Use this property to make the optional field visible on the Input section. For Ex: Since is optional field but to be displayed always under Input fields. Use <code>stick: :true</code>.
+            <td>Use this property to make the optional field visible on the Input section. For Ex: Since is optional field but to be displayed always under Input fields. Use <code>sticky: true</code>.
             </td>
         </tr>
         <tr>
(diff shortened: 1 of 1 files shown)


developing-connectors/sdk/object-definition.md
```diff
@@ -67,6 +67,16 @@ object_definitions: {
             </td>
         </tr>
         <tr>
+            <td>label</td>
+            <td>The Title of this field.
```
eeshan92: Let's change this to: "An optional key. All fields will have default labels based on the field name. Use this to change the default value of the field label."

developing-connectors/sdk/object-definition.md
```diff
@@ -67,6 +67,16 @@ object_definitions: {
...
+            </td>
+        </tr>
+        <tr>
+            <td>stick</td>
```
eeshan92: Should be `sticky`

developing-connectors/sdk/object-definition.md
```diff
@@ -67,6 +67,16 @@ object_definitions: {
...
+            <td>Use this property to make the optional field visible on the Input section. For Ex: Since is optional field but to be displayed always under Input fields. Use <code>sticky: true</code>.
```
eeshan92: Rephrase to:

Boolean field used to force a field to always be visible by default  as input fields, even optional fields. Example: From (since) input is an optional trigger input field but should always be displayed. Default value is <code>false</code>.

developing-connectors/sdk/object-definition.md
```diff
@@ -67,6 +67,16 @@ object_definitions: {
             </td>
         </tr>
         <tr>
+            <td>label</td>
+            <td>An optional key. All fields will have default labels based on the field name. Use this to change the default value of the field label.
+            </td>
+        </tr>
+        <tr>
+            <td>sticky</td>
```
eeshan92: Last change, can oyu move sticky to the bottom on the table? It is not important enough as a key that needs to be 3rd from the top.

Issue #250, opened 2017-09-12
More details for OAuth 2.0 documentation


eeshan92:
- [ ] what happens in the back after user authorizes
- [ ] how to handle other variants (Create a new file called Custom OAuth 2.0)

PR #257, opened 2017-09-20
Add custom token OAuth2 into SDK connector docs

njyjn:
## Update

- Moved OAuth2 introduction to the top of the document.
- Information for custom token auth for OAuth2:
  - Using the `acquire` hook in place of the `token_url`
- `refresh_on`, `detect_on` and `refresh` information (largely imported from custom auth section) included within the same page.
  - I recommend we create a new section on refresh flow since both OAuth2 and custom auth use them. Thoughts?


developing-connectors/sdk/authentication/oauth2-authentication.md
```diff
@@ -74,7 +76,176 @@ connection: {
...
+  "settings": "no"
+}
+```
+we can retrieve the `access_token` from `response[access_token]`. These parameters are also appended into the original `connection` object.
+
+The `acquire` hook must return an array of
+
+- Hash: Access token, [Refresh token](#refresh-tokens) (optional)
+- Owner ID for clobber detection (optional, substitute with `nil`)
+- Hash: Other settings to merge (optional, substitute with `nil`)
```
eeshan92: @njyjn Updated with more details

---
When using custom OAuth2 type connection, the `acquire` hook must return an array with the following values in sequence:
- Tokens
- Owner ID
- Other values

#### Tokens
Needs to be a hash with the exact key name (`access_token`, [`refresh_token`](#refresh-tokens). If the API returns tokens with other keys (example: `id_access` and `id_refresh` respectively), you can map them here.
```ruby
{
  "access_token": response["id_access"],
  "refresh_token": response["id_refresh"]
}
```
#### Owner ID
Optional value used for clobber detection (substitute with `nil` if not used)

#### Other values
Optional hash that will be merged with the **original connection hash**. Assuming you have this original connection hash.
```ruby
{
  client_id: "CLIENT_ID",
  client_secret: "CLIENT_SECRET"
}
```

and you wish to merge the `company_id` value from the response, you can pass it here with the following hash.
```ruby
{ company_id: response["company_id"] }
```

The resulting connection hash will look like this.
```ruby
{
  client_id: "CLIENT_ID",
  client_secret: "CLIENT_SECRET",
  company_id: response["company_id"]
}
```
In the rest of the custom adapter, this value can be referenced with `connection["company_id"]`

njyjn:
Should resolve #250

PR #262, opened 2017-09-21
workbot new command trigger writeup

huilinn:


workbot/workbot-triggers.md
```diff
@@ -4,52 +4,111 @@ date: 2017-03-30 05:00:00 Z
...
+| Business data                   | 3rd part of the Workbot command. The record that the Workbot command works with, e.g. invoice, customer, ticket.                                                                                                                                                                                                                                                                                                                                                           |
+| Input parameters                | The data to ask from the user so as to successfully carry out recipe actions, defined in a specific format.  Parameters may have the type *file*, which takes as input file content (e.g. text, binary content), that you may upload into Slack. You can use this file in subsequent operations using the Workbot action **[Download Attachment](workbot-actions.md#download-attachment)** |
```
senrong: We should also mention that `string` is default and will return parameters data in text. Not particularly sure what `date_time` does but it should parse input into date_time type. There's also a `options` type that provides input in button form that users can click to respond. 

`optional` determines whether a parameter shows up to the user by default. Users can choose to enter the parameter if this is false, but will not be prompted by default when the command start. To prompt a user to enter optional parameters, set `prompt`to true for that parameter. 

`hint` will give the user an idea of what data is expected in the parameter, and `sample` is an example of a valid value for that parameter. Both will appear beside the parameter name when thr trigger flow starts.

PR #264, opened 2017-09-26
Edited Slack workbot articles

huilinn:


workbot/workbot-triggers.md
```diff
@@ -32,6 +33,49 @@ The following table lists the configurable input fields in the trigger, and what
...
+```
+param_name optional:true prompt:false type:string hint:hint_to_help_users sample:example_data
+```
+
+| Input parameter field | Description                                                                                                    |
+|-----------------------|----------------------------------------------------------------------------------------------------------------|
+| Name                  | Required. Needs to be a single word.                                                                           |
+| Optional              | Not required. If not defined, defaults to `optional:true`. Input parameter will be optional.                   |
+| Prompt                | Not required. If not defined, defaults to `prompt:false`. No prompt will be generated.                         |
+| Type                  | Not required. If not defined, defaults to `type:string`. Possible types are `string`, `file`, and `date_time`. |
```
eeshan92: There is also `number` and `boolean` types

Issue #309, opened 2017-11-29
Improvements to action documentation


njyjn:
This will reduce confusion over when and when not to use `{ }` within the execute block.

PR #310, opened 2017-11-29
Add information on output of action's execute block

njyjn:
Added
---

### Output

Did you notice that some requests were encapsulated in `{ }`, and some were not? This will depend on what your endpoint returns and how you would like to map the response to your `output_definition`. Typically, most REST endpoints would return a JSON response, whether the request was successful or not.

Here are two examples. In all cases, the SDK expects the last line of the execute block to be a Hash object.

Assume that the output from an example endpoint is as so

```json
{
  "email": "j@a.co"
}
```

#### 1. Simple object output

Here, we can leave the `GET` method alone, since it maps directly into our `output_fields` definition.

```ruby
execute: lambda do |connection, input|
  get("https://api.example.com/v1/email")
end,

output_fields: lambda do
  [
    { name: "email" }
  ]
end
```

#### 2. Nested object output

Here we need to take the extra step of mapping the response returned as a nested Hash object.

```ruby
execute: lambda do |connection, input|
  response = get("https://api.example.com/v1/email")

  { user: response }
end,

output_fields: lambda do
  [
    { name: "user", type: :object, properties: [
      { name: "email" }
    ] }
  ]
end
```

or 

```ruby
execute: lambda do |connection, input|
  {
    user: get("https://api.example.com/v1/email")
  }
end,

output_fields: lambda do
  [
    { name: "user", type: :object, properties: [
      { name: "email" }
    ] }
  ]
end
```

Do also take precautions if the endpoint returns a non-JSON response, e.g. `"j@a.co"`. If you do not encapsulate these into a final Hash object, you may get this commonly seen error:

```
undefined method `to_hash` for "j@a.co":String
Did you mean? to_s
```

Use the method `response_format_raw` in these cases

```ruby
execute: lambda do |connection, input|
  response = get("https://api.example.com/v1/email").
    response_format_raw

    { email: response }
end,

output_fields: lambda do
  [
    { name: "email" }
  ]
end
```


njyjn:
Resolves #309

Branch built on top of #257 before fffe99d. Will resolve when #257 is merged.

developing-connectors/sdk/action.md
```diff
@@ -73,3 +73,96 @@ end
...
+Here we need to take the extra step of mapping the response returned as a nested Hash object.
+
+```ruby
+execute: lambda do |connection, input|
+  response = get("https://api.example.com/v1/email")
+
+  { user: response }
+end,
+
+output_fields: lambda do
```
eeshan92: Let's make the definition clearer by separating the lines
```ruby
  [
    {
      name: "user",
      type: :object,
      properties: [
        { name: "email" }
      ]
    }
 ]
```

developing-connectors/sdk/action.md
```diff
@@ -73,3 +73,96 @@ end
...
+
+    { email: response }
+end,
+
+output_fields: lambda do
+  [
+    { name: "email" }
+  ]
+end
+```
```
eeshan92: can we also add example for top level array responses

developing-connectors/sdk/authentication/oauth2-authentication.md
```diff
@@ -74,7 +76,232 @@ connection: {
...
+- The `token_url` is required if the `acquire` or `refresh` hooks are not present (see below).
 - Ensure that your implementation of OAuth 2.0 is compliant with the specifications stated in the RFC document. Else, your custom adapter might not start.
   - For example, related to [Issuing an Access Token - Successful Response](https://tools.ietf.org/html/rfc6749#section-5.1), Workato will be expecting a response with `access_token` in the response. Returning the access token under a key of `accessToken` in a JSON response will result in an unsuccessful Workato request to your `token_url`.
   - Usually this will not be a problem because most OAuth libraries out there will do most of the heavy lifting for you, such as returning response in the right format etc. It is good to be aware of this!
+
+### Custom token authentication
+The `token_url` is called using a `POST` request with the provided `client_id` and `client_secret` appended into the header.
+
+However, some APIs require the authorization token to be obtained by using a `POST` request with basic authentication. Use the `acquire` hook instead.
```
eeshan92: Mention that we will show an example with basic. Right now, it looks like the only custom auth we support is basic auth using client ID/secret.

developing-connectors/sdk/authentication/oauth2-authentication.md
```diff
@@ -74,7 +76,232 @@ connection: {
...
+      apply: lambda do |connection, access_token|
+        headers("Authorization": "Bearer #{connection["access_token"]}")
+      end
+    }
+  }
+```
+
+The methods `.user` and `.password` are the equivalent of appending `Authorization: BASIC ` and `<user>:<password>` in BASE-64 String encoding in the `POST` request header. Note that the request must be sent with `request_format_www_form_urlencoded`.
+
+Upon receiving a the request, the API returns a JSON response. These can be accessed using the `response[key...]` variable defined. For example, if the call returns
```
eeshan92: typo "a the"

eeshan92: replace `response[key...]` with  `response["key"]`

developing-connectors/sdk/authentication/oauth2-authentication.md
```diff
@@ -74,7 +76,232 @@ connection: {
...
+
+```json
+{
+  "access_token": 12345,
+  "refresh_token": 12345,
+  "settings": "no"
+}
+```
+we can retrieve the `access_token` from `response[access_token]`. These parameters are also appended into the original `connection` object.
```
eeshan92: replace `response[access_token]` with `response["access_token"]`

eeshan92: "These parameters..." are not automatically merged into connection hash. Let's remove this sentence to avoid confusion.

developing-connectors/sdk/action.md
eeshansim: What if the response is an array of emails. Our practice is the wrap the array in a hash envelope.

{ emails: response }

njyjn:
@eeshansim 

## Added 

#### 3. Array output

##### Top-level array responses

Some endpoints return a simple top-level array response

```json
[
  "elijah@a.co",
  "marion@a.co",
  "shayna@a.co"
]
```

We would simply need to nest the object like in the previous example

```ruby
execute: lambda do |connection, input|
  response = get("https://api.example.com/v1/emails")

    { emails: response }
end,

output_fields: lambda do
  [
    {
      name: "emails",
      type: :array,
      of: :string
    }
  ]
end
```

##### Nested array responses

Now let's assume that the output from an example endpoint is as so

```json
{
  "child_emails": [
    "elijah@a.co",
    "marion@a.co",
    "shayna@a.co"
  ]
}
```

We can directly map the response in `output_fields` without having to nest it.

```ruby
execute: lambda do |connection, input|
  get("https://api.example.com/v1/child_emails")
end,

output_fields: lambda do
  [
    {
      name: "child_emails",
      type: :array,
      of: :string
    }
  ]
end
```

Note that we define an array of name `child_emails` in our `output_fields`. This matches the response array's name. If you intend to use different names, you must use nest the response output in the `execute` block.

For example, we want the output array to have name `emails` instead of `child_emails`. Let's use Ruby method `dig` to surface the desired array to the top-level.

```ruby
execute: lambda do |connection, input|
  response = get("https://api.example.com/v1/child_emails").
               dig["child_emails"]
               # ["elijah@a.co","marion@a.co","shayna@a.co"]

  { emails: response }
end,

output_fields: lambda do
  [
    {
      name: "emails",
      type: :array,
      of: :string
    }
  ]
end
```

##### Responses with arrays of objects

```json
{
  "children": [
    {
      "child_name": "Elijah",
      "child_age": 3
    },
    {
      "child_name": "Marion",
      "child_age": 1
    }
  ]
}
```
The only thing that is different would be how we define the output. Let's use a predefined object in `object_definitions`, which is similar to the nomenclature used in the response.

```ruby
object_definitions: {
  children: {
    fields: lambda do
      [
        {
          name: "children",
          type: :array,
          of: :object,
          properties: [
            { name: "child_name" },
            { name: "child_age", type: :integer }
          ]
        }
      ]
  }
}
```

With this, we can simply perform the mapping in the following way

```ruby
execute: lambda do |connection, input|
  get("https://api.example.com/v1/children")
end,

output_fields: lambda do
  object_definitions["children"]
end
```

developing-connectors/sdk/action.md
```diff
@@ -73,3 +73,246 @@ end
 ```
 
 See [Methods](/developing-connectors/sdk/methods.md) section for list of methods available for use in your custom connector actions.
+
+### Output
+
+Did you notice that some requests were encapsulated in `{ }`, and some were not? This will depend on what your endpoint returns and how you would like to map the response to your `output_definition`. Typically, most REST endpoints would return a JSON response, whether the request was successful or not.
```
eeshansim: is `output_definition` a typo of `output_fields` definition?

developing-connectors/sdk/action.md
```diff
@@ -73,3 +73,246 @@ end
...
+
+Assume that the output from an example endpoint is as so
+
+```json
+{
+  "email": "j@a.co"
+}
+```
+
+#### 1. Simple object output
```
eeshansim: we don't usually number our titles

developing-connectors/sdk/action.md
```diff
@@ -73,3 +73,246 @@ end
...
+
+Here, we can leave the `GET` method alone, since it maps directly into our `output_fields` definition.
```
eeshansim: Good if we can show an example of what the expected response looks like

developing-connectors/sdk/action.md
```diff
@@ -73,3 +73,246 @@ end
...
+
+#### 2. Nested object output
+
+Here we need to take the extra step of mapping the response returned as a nested Hash object.
+
+```ruby
+execute: lambda do |connection, input|
+  response = get("https://api.example.com/v1/email")
+
+  { user: response }
```
eeshansim: not sure why we need this example. The "nesting" is created artificially in `execute` which makes the exatra layer in `output_fields` redundant. This example if necessary only if the response is already nested, like this:

```json
{
  "user": {
    "email": "john@acme.corp"
  }
}
```

developing-connectors/sdk/action.md
```diff
@@ -73,3 +73,246 @@ end
...
+```
+
+Do also take precautions if the endpoint returns a non-JSON response, e.g. `"j@a.co"`. If you do not encapsulate these into a final Hash object, you may get this commonly seen error:
+
+```
+undefined method `to_hash` for "j@a.co":String
+Did you mean? to_s
+```
+
+Use the method `response_format_raw` in these cases
```
eeshansim: Need to be more explicit. Something like this:
> Also, we need to explicitly tell the request to expect a non-JSON response. In this case, expect `response_format_raw`. This will skip the default auto JSON parser and prevent an error shown above.

developing-connectors/sdk/action.md
njyjn: Maybe a screenshot of the recipe pill structure?

njyjn: Oh or did you mean 

> Assume that the output from an example endpoint is as so

```json
{
  "email": "j@a.co"
}
```

developing-connectors/sdk/action.md
njyjn: What if we needed to store the email into a user object?

developing-connectors/sdk/action.md
eeshansim: Yes the second one

PR #326, opened 2017-12-30
Kafka configuration

spn:
This might be too technical. Probably could be made less verbose.


huilinn:
I think it's okay, it's not super long. @kandadaboggu can you approve this too before merging, as you have more context about the kafka connector?

PR #328, opened 2018-01-02
Add information on reusable methods

njyjn:
As requested, here is the documentation on reusable methods. It covers

1. General information on use
2. Notes on the method's input object, recursion, and testing.

and includes a link from`methods.md`

### DIFF
--- a/developing-connectors/sdk/reusable-methods.md
+++ b/developing-connectors/sdk/reusable-methods.md
@@ -1,0 +1,94 @@
+# Reusable Methods
+
+Reusable methods are now supported in Workato. This is nifty if you seek to reuse custom methods anywhere within an executable lambda (e.g. `execute`, `object_definitions/fields`).
+
+Such methods are declared using the `method` block. This block lies in the same level as the `trigger` and `action` blocks.
+
+## Example
+```ruby
+methods: {
+  factorial: lambda do |input|
+    number = input[:number]
+    if number > 1
+      number * call(:factorial, { number: number - 1 })
+    else
+      number
+    end
+  end
+
+  hello: lambda do
+    puts "Hello world"
+  end
+},
+
+actions: {
+  factorial: {
+    execute: lambda do |connection, input|
+      { factorial: call(:factorial, { number: input["number"] }) }
+    end
+  },
+
+  say_hello: {
+    execute: lambda do
+      call(:hello)
+    end
+  }
+},
+```
+
+## The `call` method
+
+```ruby
+call(:name, { number: 1 })
+```
+
+Use the `call()` method to reference a method. This method takes in two parameters:
+
+1. Method name
+  - Use the method name defined. You can use either `:name` or `"name"` representations.
+2. Input hash (optional)
+  - This is a multi-valued hash of input variables. Leave blank if your method does not take in an input.
+
+The input hash is later appended to the main `input` object, which you can reference as so
+
+```ruby
+factorial: lambda do |input|
+  number = input[:number]
+...
+```
+
+This means that within your method, you may also make reference to any other keys stored in the global `input` hash. However, you may not reference any local input variables outside of this lambda.
+
+The following code block works where `input[:constant]` was defined by the user at runtime.
+```ruby
+factorial_extension: lambda do |input|
+  number = input[:number] + input[:constant]
+end
+```
+
+If `number` was an input variable to the `factorial` method, the following code block will not work.
+```ruby
+test_action: {
+  execute: lambda do |connection, input|
+    call(:factorial, { number: 1 } )
+    puts input[:number]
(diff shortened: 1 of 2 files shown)


developing-connectors/sdk/reusable-methods.md
```diff
@@ -0,0 +1,94 @@
+# Reusable Methods
+
+Reusable methods are now supported in Workato. This is nifty if you seek to reuse custom methods anywhere within an executable lambda (e.g. `execute`, `object_definitions/fields`).
```
eeshan92: 1. No need to say supported now
2. we should emphasise the [DRY](https://www.codementor.io/joshuaaroke/dry-code-vs-wet-code-89xjwv11w) principle
3. methods can be reused in all code blocks, not just execute (apply, fields, pick_lists etc)

Let's rephrase to something the includes "reusable methods help to keep your custom adapter code DRY" and "can be used in any code block"

developing-connectors/sdk/reusable-methods.md
```diff
@@ -0,0 +1,94 @@
...
+
+Such methods are declared using the `method` block. This block lies in the same level as the `trigger` and `action` blocks.
```
eeshan92: better way to say is that it is a top level key in the custom adapter code, similar to `triggers` and `actions`

developing-connectors/sdk/reusable-methods.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Example
+```ruby
```
eeshan92: since this is the first and only example with methods, let's include the top level hash to reduce confusion

developing-connectors/sdk/reusable-methods.md
```diff
@@ -0,0 +1,94 @@
...
+
+  hello: lambda do
+    puts "Hello world"
+  end
+},
+
+actions: {
+  factorial: {
+    execute: lambda do |connection, input|
+      { factorial: call(:factorial, { number: input["number"] }) }
```
eeshan92: since no input fields are defined, it can be confusing to a reader. Can we either add input fields are just use a hardcoded number

developing-connectors/sdk/reusable-methods.md
```diff
@@ -0,0 +1,94 @@
...
+## The `call` method
+
+```ruby
+call(:name, { number: 1 })
+```
+
+Use the `call()` method to reference a method. This method takes in two parameters:
+
+1. Method name
+  - Use the method name defined. You can use either `:name` or `"name"` representations.
```
eeshan92: we should explain what these are
"`:method_name (symbol)` or `"method_name" (string)` representations"

developing-connectors/sdk/reusable-methods.md
```diff
@@ -0,0 +1,94 @@
...
+2. Input hash (optional)
+  - This is a multi-valued hash of input variables. Leave blank if your method does not take in an input.
```
eeshan92: what is multi-valued?

developing-connectors/sdk/reusable-methods.md
```diff
@@ -0,0 +1,94 @@
...
+
+The input hash is later appended to the main `input` object, which you can reference as so
```
eeshan92: not appended, it is passed as an argument

developing-connectors/sdk/reusable-methods.md
```diff
@@ -0,0 +1,94 @@
...
+
+```ruby
+factorial: lambda do |input|
+  number = input[:number]
+...
+```
+
+This means that within your method, you may also make reference to any other keys stored in the global `input` hash. However, you may not reference any local input variables outside of this lambda.
```
eeshan92: I think there is a misunderstanding here. This `input` is not a reference to the input from which it was called.

developing-connectors/sdk/reusable-methods.md
```diff
@@ -0,0 +1,94 @@
...
+
+The following code block works where `input[:constant]` was defined by the user at runtime.
```
eeshan92: See above

developing-connectors/sdk/reusable-methods.md
```diff
@@ -0,0 +1,94 @@
...
+test_action: {
+  execute: lambda do |connection, input|
+    call(:factorial, { number: 1 } )
+    puts input[:number]
+  end
+```
+
+
+## Recursion
+In the case of the factorial example provided above, note that the `factorial` method has some degree of recursion. Be careful when writing recursive loops by setting failsafes as in the example below:
```
eeshan92: Can we add
"methods can also be called within method code blocks. This means that a method can be called by another method or by itself. In the case of the..."

PR #333, opened 2018-01-05
Edit OPA installation guide & screenshots

zweedao:
For this to work, the other pull request needs to be approved first: https://github.com/workato/docs/pull/332

### DIFF
--- a/on-prem.md
+++ b/on-prem.md
@@ -56,6 +56,7 @@ Please make sure that TCP port 3000 is available for binding.
 3. [Start agent](/on-prem.md#start-agent)
 
 ## Install agent
+### For Windows
 1. On the top navigation bar, click `Tools` > `On-prem agent`
 
 ![On-prem option](assets/images/on-prem/on_prem_access_option_Jan_2018.png)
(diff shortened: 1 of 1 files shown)


on-prem.md
```diff
@@ -10,9 +10,9 @@ The Workato on-premise agent provides a secure way for Workato to selectively ac
 
 On-premise access is enabled only for certain plans. Check the [Pricing and Plans page](https://www.workato.com/pricing?audience=general) or reach out to Workato sales representatives at +1 (844) 469-6752 to find out more.
 
-For Workato accounts with on-prem access, users are able to view the on-prem option on their account menu.
+For Workato accounts with on-prem access, users are able to view the on-prem option in the Tools menu.
 
-![On-prem option](/assets/images/on-prem/on_prem_access_option.png)
+![On-prem option](assets/images/on-prem/on_prem_access_option_Jan_2018.png)
```
huilinn: @zweedao let's put the screenshots in the same PR as the changes next time, so that we can preview the PR to see whether the images are displayed correctly

on-prem.md
```diff
@@ -56,20 +56,35 @@ Please make sure that TCP port 3000 is available for binding.
 3. [Start agent](/on-prem.md#start-agent)
 
 ## Install agent
```
huilinn: @zweedao do you think it'd be better if we have 2 different sections for `##Install agent on Windows` and `##Install agent on Linux and Mac OS`? Each section should have their own steps 1-6 and steps 1-8 respectively. That way it's easier for users to follow, and subsequently even if the installation instructions changes to have no steps in common, we don't need to make backward incompatible link changes and break links pointing to these sections.

on-prem.md
zweedao: ok noted

on-prem.md
zweedao: I think repeating step 1-4 is quite long. I have added line 75 to make it clearer that there will be next steps based on different operating systems.

PR #336, opened 2018-01-08
Added nested where formula eg

huilinn:
https://github.com/workato/issues/issues/5580


formulas/array-list-formulas.md
```diff
@@ -154,8 +153,23 @@ These rows will be expressed as a list of hashes:
 ]
 ```
 
+The following is a simple WHERE formula retrieving lead records with the company Delphi Chemicals.
+![Simple where formula retrieving lead records from company Delphi Chemicals](/assets/images/formula-docs/non-nested-where-formula.png)
+*Simple where formula retrieving lead records from company Delphi Chemicals*
+
+### Example of nested where formula
```
eeshan92: @huilinn more precise phrasing is where formula for nested fields. Saying nested where may cause users to think of using a where formula within another one. What do you think?

formulas/array-list-formulas.md
```diff
@@ -154,8 +153,23 @@ These rows will be expressed as a list of hashes:
...
+The WHERE formula can be nested to filter for records by values not at the top level of the datatree.
```
eeshan92: Same here, we should explain that this section is to describe querying arrays of hashes based on a nested field

formulas/array-list-formulas.md
huilinn: Sure, do you have any thoughts on what's a more precise term?

eeshan92: how about:
where formula for nested fields

huilinn:
@eeshan92 have fixed comments

PR #361, opened 2018-02-07
DB docs template

eeshansim:
Connector
- [x] connection setup
- [x]  Single vs Batch actions
- [x] Using the `WHERE` condition

Actions
- [x] Select actions
- [x] Insert actions
- [x] Update actions
- [x] Upsert actions
- [x] Delete actions
- [ ]  Execute stored procedures

Triggers
- [ ] New row triggers
- [ ] New/update row triggers
- [ ] Scheduled query


connectors/mysql.md
```diff
@@ -0,0 +1,97 @@
...
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td>Connection name</td>
+      <td>Give this MySQL connection a unique name that identifies which MySQL instance it is connected to.</td>
+    </tr>
+    <tr>
+      <td>On-prem secure agent</td>
+      <td>Choose an on-premise agent if your database is running in a network that does not allow direct connection. Before attempting to connect, make sure you have an active on-premise agent. Refer to the [On-premise agent](/on-prem.md) guide for more information.</td>
```
huilinn: This looks like it's not rendering well in the table:
screen shot 2018-02-28 at 4 05 41 pm

connectors/mysql.md
```diff
@@ -0,0 +1,97 @@
...
+      <td>Database</td>
+      <td>Name of the MySQL database you wish to connect to.</td>
+    </tr>
+  </tbody>
+</table>
+
+## Working with the MySQL connector
+
+### Table, view and stored procedure
+MySQL connector works with all tables, views and stored procedures. These are available in pick lists in each trigger/action or you can provide the exact name.
```
huilinn: Can we mention if the exact name is case sensitive or not?

huilinn:
@eeshansim I think it would be nice to have:
* [ ] example recipes for single rows VS batches
* [ ] screenshots of the datatree pointing out when an action returns a list VS when an action returns a row
* [ ] screenshots of actual JSON output for list VS rows, and explanation of when we summarize the data (e.g. I think for batches, we don't always show all JSON output?) would be good, but maybe too much

We can put these in a separate issue if that's easier.

PR #371, opened 2018-02-14
368 formula mode update

### LINKED ISSUE #368: formula mode update
- [ ] broken nav
links nested under formula mode are not displayed
image

- [ ] complex data types

eeshansim:
closes #368


formulas/formula-mode.md
```diff
@@ -0,0 +1,90 @@
+---
```
huilinn: @eeshansim, instead of creating a new article, can we stick to the same format? We would be breaking links that point to the old article from the product, otherwise.

formulas/complex-data-types.md
```diff
@@ -0,0 +1,45 @@
+---
```
huilinn: @eeshansim, this use case seems like something we should put in arrays. Alternatively, can we name this something more descriptive like JSON object formulas?

formulas/formula-mode.md
eeshansim: @huilinn reason for this is that the existing format does not work.

formula-mode-link

I have checked all existing links in the product and it does not affect any.

formulas/complex-data-types.md
eeshansim: My initial thought was this:
Data types are generally split into 2, primitive (date, string, number) and complex (array, object). List formulas deal with arrays, but ultimate convert them into strings/numbers for mapping. This feels like a separate class (maintaining object/array structure in mapping). Does it still make sense to name is JSON object formulas?

How about "Advanced formulas"? Since this is not exactly mapping field values into an input field (mapping arrays and objects into fields)

FYI: Have not added JSON objects in the examples yet because I'm still thinking of a valid use case.

eeshansim:
@huilinn responded

PR #382, opened 2018-03-06
Improve DB docs template

### LINKED ISSUE #380: MySQL/DB template enhancements
* [ ] add link to how-to create connection profiles on OPA
* [x] some more basic examples of how to build simple SQL queries on MySQL in the WHERE section, e.g. conventions with double quotes, single quotes, common data types, common operators, or maybe just a link to MySQL syntax docs?
* [x] example recipes for single rows VS batches
* [x] screenshots of the datatree pointing out when an action returns a list VS when an action returns a row
* [ ] screenshots of actual JSON output for list VS rows, and explanation of when we summarize the data (e.g. I think for batches, we don't always show all JSON output?) would be good, but maybe too much

eeshansim:
fixes #380


connectors/mysql.md
```diff
@@ -71,21 +71,69 @@ MySQL connector can read or write to your database either as a single row or in
 ![Batch trigger inputs](/assets/images/mysql/batch_trigger_input.png)
 *Batch trigger inputs*
 
+Besides the difference in input fields, there is also difference between the outputs of these 2 types of operations. A trigger that process rows one at a time will have an output datatree that allows you to map data from that single row.
```
jandonyada: Minor grammatical error "...there is also **a** difference..."

connectors/mysql.md
```diff
@@ -71,21 +71,69 @@ MySQL connector can read or write to your database either as a single row or in
...
+String values must be enclosed in single quotes (`''`) and columns used must exist in the table.
 
 A simple `WHERE` condition to filter rows based on values in a single column looks like this.
 
 ```sql
 currency = 'USD'
 ```
 
-If used in a **Select rows** action, this `WHERE` condition will return all rows that has the value 'USD' in the `currency` column.
+If used in a **Select rows** action, this `WHERE` condition will return all rows that has the value 'USD' in the `currency` column. Just remember to wrap datapills with single quotes in your inputs.
```
jandonyada: "this `WHERE` condition will return all rows that **have** the value"

connectors/mysql.md
jandonyada: Minor grammatical errors 
- "...there is also **a** difference..."
- "A trigger that **processes**"

connectors/mysql.md
```diff
@@ -71,21 +71,69 @@ MySQL connector can read or write to your database either as a single row or in
...
+```
+
+In a recipe, remember to add backticks to the column identifiers.
+
+![Using datapills in WHERE condition with backticks](/assets/images/mysql/use_datapill_in_where_backtick.png)
+*Using datapills in `WHERE` condition backticks*
+
+Double quotes (`""`) can also be used for string values but is less commonly accepted in other databases. For this reason, single quotes are used more widely than double quotes.
+
+MySQL also expects `DATE` and `DATETIME` values be single quoted. You can use double quotes for other column types.
```
jandonyada: "MySQL also expects `DATE` and `DATETIME` values **to** be single quoted."

connectors/mysql.md
```diff
@@ -71,21 +71,69 @@ MySQL connector can read or write to your database either as a single row or in
...
+
+```sql
+created_date > '2018-03-01' and currency = "USD"
+```
+
+In a recipe, remember to use the appropriate quotes for each values.
```
jandonyada: In a recipe, remember to use the appropriate quotes for each value**~~s~~**.

connectors/mysql.md
```diff
@@ -71,21 +71,69 @@ MySQL connector can read or write to your database either as a single row or in
...
+
+Backticks (` `` `) in `WHERE` statements are for tables and columns identifiers. This is required when the identifier is a MySQL reserved keyword or contains special characters.
+
+```sql
+`currency` = 'USD'
+```
+
+In a recipe, remember to add backticks to the column identifiers.
+
+![Using datapills in WHERE condition with backticks](/assets/images/mysql/use_datapill_in_where_backtick.png)
```
jandonyada: Should rename to /use_datapill_in_where_backtick**s**.png for consistency with the text

PR #433, opened 2018-04-22
Stored proc and custom SQL actions

### LINKED ISSUE #431: DB: Stored procedure and custom SQL action

eeshansim:
fixes #431


connectors/mssql/select.md
```diff
@@ -43,3 +43,19 @@ This input field determines the maximum number of rows to return. The default li
 
 ### Offset
 This input field gives you the option to fetch only a page of results from the entire results set. For example, to skip the first 100 rows of the selected results set, input `100` to this field. The default is `0`.
+
+## Select rows using custom SQL
+This action lets you select rows based on a custom SQL query. Rows that are returned from the SQL will be returned as the output of this action.
```
jandonyada: Rows that are returned from the **SQL query**

connectors/mssql/select.md
```diff
@@ -43,3 +43,19 @@ This input field determines the maximum number of rows to return. The default li
...
+
+![Select rows using custom SQL action](/assets/images/mssql/custom-sql-action.png)
+*Select rows using custom SQL action*
+
+### SQL
+Provide the SQL to be executed to select rows. The SQL here will be used to generate the output datatree. To do this, the SQL will be executed once when you provide it. You can map datapills here to execute dynamically changing SQL statements. Remember to wrap datapills in quotes (`''`).
+
+Avoid using limit clause like `TOP` in your SQL. This is because a limit to the size of rows returned in the SQL based on the value defined in the **Limit** input field. Adding your own limit clause will cause the SQL to fail.
```
jandonyada: Avoid using limit clause**s** like `TOP`

jandonyada: > This is because a limit to the size of rows returned in the SQL based on the value defined in the **Limit** input field

Did you mean:
This is because ~~a~~ **the** limit to the size of rows returned in the SQL **query is** based on the value defined in the **Limit** input field (more on this [below](/connectors/myssql/select.md#limit)). <-- how about this for clarity?

Adding your own limit clause will cause the SQL **query** to fail.

connectors/mssql/select.md
jandonyada: At the risk of sounding like a broken record (since SQL query comes up pretty often in the doc), we could just say 'query' instead.

connectors/mssql/stored-procedure.md
```diff
@@ -0,0 +1,21 @@
+---
+title: Workato connectors - SQL Server Execute stored procedure actions
+date: 2018-03-19 06:00:00 Z
+---
+
+# SQL Server - Execute stored procedure action
+
+## Execute stored procedure
+This action lets you executed any stored procedure or functions saved in your database instance.
```
jandonyada: This action lets you execute~~d~~

connectors/oracle/select.md
```diff
@@ -43,3 +43,19 @@ This input field determines the maximum number of rows to return. The default li
 
 ### Offset
 This input field gives you the option to fetch only a page of results from the entire results set. For example, to skip the first 100 rows of the selected results set, input `100` to this field. The default is `0`.
+
+## Select rows using custom SQL
+This action lets you select rows based on a custom SQL query. Rows that are returned from the SQL will be returned as the output of this action.
```
jandonyada: Same here, SQL **query**

connectors/oracle/select.md
```diff
@@ -43,3 +43,19 @@ This input field determines the maximum number of rows to return. The default li
...
+
+![Select rows using custom SQL action](/assets/images/oracle/custom-sql-action.png)
+*Select rows using custom SQL action*
+
+### SQL
+Provide the SQL to be executed to select rows. The SQL here will be used to generate the output datatree. To do this, the SQL will be executed once when you provide it. You can map datapills here to execute dynamically changing SQL statements. Remember to wrap datapills in quotes (`''`).
+
+Avoid using limit clause like `ROWNUM` in your SQL. This is because a limit to the size of rows returned in the SQL based on the value defined in the **Limit** input field. Adding your own limit clause will cause the SQL to fail.
```
jandonyada: > Avoid using limit clause like `ROWNUM` in your SQL. This is because a limit to the size of rows returned in the SQL based on the value defined in the **Limit** input field. Adding your own limit clause will cause the SQL to fail.

Did you mean:
Avoid using limit clause**s** like `ROWNUM` in your SQL. This is because ~~a~~ **the** limit to the size of rows returned in the SQL **is** based on the value defined in the **Limit** input field (more on this [below](/connectors/oracle/select.md#limit)).  <-- how about this for greater clarity?

Adding your own limit clause will cause the SQL **query** to fail.

connectors/oracle/stored-procedure.md
```diff
@@ -0,0 +1,21 @@
+---
+title: Workato connectors - Oracle Execute stored procedure actions
+date: 2018-03-19 06:00:00 Z
+---
+
+# Oracle - Execute stored procedure action
+
+## Execute stored procedure
+This action lets you executed any stored procedure or functions saved in your database instance. To use this action, your Oracle connection must be established using an [On-premise agent](/on-prem.md). Stored procedures from packages are not supported in this action.
```
jandonyada: execute~~d~~

connectors/mssql/select.md
eeshansim: That's a good point.

PR #447, opened 2018-04-27
NTLM Profile section added

bmvkrd:


bmvkrd:
Huilin, please review the change

on-prem.md
```diff
@@ -270,6 +270,28 @@ kafka:
...
+```
+
+The following properties are supported:
+
+| Property name | Description |
+|------------------|-------------------------------------------|
+| auth | NTLM authentication credentials |
+| base_url | The base URL for NTLM resources |
+| cm_default_max_per_route | (Optional) Sets the number of connections per route (default 5) |
+| cm_max_total | (Optional) Sets the maximum number of connections (default 10) |
```
eeshansim: Is there a max or min number?

on-prem.md
```diff
@@ -270,6 +270,28 @@ kafka:
...
+    cm_max_total: 100
+```
+
+The following properties are supported:
+
+| Property name | Description |
+|------------------|-------------------------------------------|
+| auth | NTLM authentication credentials |
+| base_url | The base URL for NTLM resources |
+| cm_default_max_per_route | (Optional) Sets the number of connections per route (default 5) |
```
eeshansim: Is there a max or min number?

eeshansim: Also, will be good be to define what our definition of a route is.

on-prem.md
bmvkrd: Yep, set to positive

on-prem.md
bmvkrd: There is no "our" definition of a route, it is defined by HttpClient library, and quite complicated for the end users. Check "HTTP connection routing" section in this document: https://hc.apache.org/httpcomponents-client-ga/tutorial/html/connmgmt.html

PR #459, opened 2018-05-08
MySQL scheduled query trigger doc

eeshansim:
Scheduler needs redesign, but in the meantime, this doc will help with configuring.


SUMMARY.md
```diff
@@ -210,6 +210,7 @@
     * [Delete actions](connectors/mysql/delete.md)
     * [New row trigger](connectors/mysql/new-row-trigger.md)
     * [New/updated row trigger](connectors/mysql/updated-row-trigger.md)
+    * [Scheduled query trigger](connectors/mysql/scheduled-query-trigger.md)
```
jandonyada: Text cursor is in an awkward position, might wanna retake the screenshot
screen shot 2018-05-08 at 5 33 31 pm

connectors/mysql/scheduled-query-trigger.md
```diff
@@ -0,0 +1,93 @@
...
+    </tr>
+    </tr>
+    <tr>
+      <td><a href="#minute">Minute</a></td>
+      <td>Configure the minute in the hour you wish to execute the SQL query.</td>
+    </tr>
+    </tr>
+    <tr>
+      <td><a href="#day-of-month">Day of month</a></td>
+      <td>Configure the day of the moenth you wish to execute the SQL query.</td>
```
jandonyada: Typo here.

connectors/mysql/scheduled-query-trigger.md
```diff
@@ -0,0 +1,93 @@
...
+    </tr>
+    <tr>
+      <td><a href="#days-of-week">Days of week</a></td>
+      <td>Select the days of week you wish to execute the SQL query.</td>
```
jandonyada: days of **the** week

connectors/mysql/scheduled-query-trigger.md
```diff
@@ -0,0 +1,93 @@
...
+Select the timezone for the scheduler to be set in. This is a required field.
+
+### Hour
+Configure the hour of the day you wish to execute the scheduled SQL query. This is required only if you selected **daily** or **monthly** as the [Schedule](#schedule).
+
+### Minute
+Configure the minute in the hour you wish to execute the scheduled SQL query. This is always required.
+
+### Day of month
+Provide values between **1** to **31** for the day of the month will be rounded down to last day of month for shorter months. Needed only if you selected **monthly** as the [Schedule](#schedule). This is an optional field that defaults to first day of the month if left empty.
```
jandonyada: > Provide values between **1** to **31** for the day of the month will be rounded down to last day of month for shorter months.

Did you mean something like "Provide values between **1** to **31** for the day of the month. This will be rounded down to the last day of the month for shorter months."

SUMMARY.md
eeshansim: updated screenshot

connectors/mysql/scheduled-query-trigger.md
```diff
@@ -0,0 +1,95 @@
...
+Select the timezone for the scheduler to be set in. This is a required field.
+
+### Hour
+Configure the hour of the day you wish to execute the scheduled SQL query. This is required only if you selected **daily** or **monthly** as the [Schedule](#schedule).
+
+### Minute
+Configure the minute in the hour you wish to execute the scheduled SQL query. This is always required.
+
+### Day of month
+Provide a value between **1** to **31** for the day of the month. This value will be rounded down to last day of month for shorter months. (Example: if you use the value **31**, the trigger run the SQL query on **31st Mar**, **30th Apr**, **31st May**...)
```
jandonyada: This value will be rounded down to **the** last day of **the** month for shorter months. (Example: if you use the value **31**, the trigger **will** run the SQL query on **31st Mar**, **30th Apr**, **31st May**...)

But this part 

> This value will be rounded down to last day of month for shorter months....

It may require the hint being updated, but could it be better like this:
If you use the value **31** and the month ends before the 31st, the trigger will run the SQL query on the last day of the month e.g. **28th Feb**, **31st Mar**, **30th Apr**, and so on.

...if I understood correctly what the trigger does lol. What do you think? I'm ok with the way it is right now too though.

PR #495, opened 2018-05-25
Custom actions docs

huilinn:


developing-connectors/custom-actions.md
```diff
@@ -0,0 +1,200 @@
+---
+title: Custom actions
+date: 2018-05-24 14:00:00 Z
+---
+
+# Custom actions
+If you're looking to build additional actions for a connector that Workato already supports, custom actions enable you to utilize the existing connector framework instead of developing from scratch via the [HTTP universal connector](/developing-connectors/http.md) or the [SDK](/developing-connectors/sdk.md). This allows you to easily build your action by telling Workato what the action's request and response should look like, which can be obtained from the API documentation.
```
eeshansim: This is quite a long sentence with many things going on. I had difficulty reading it initially. Can we break it down into simpler sentences? Maybe into 2 sentences addressing the 2 main points:

1. Workato supports a connector to the application you are interested in, but does not support a specific action.
2. You can build a custom action on top of the existing connector framework instead of building the whole things from scratch via HTTP or the SDK.

developing-connectors/custom-actions.md
```diff
@@ -0,0 +1,200 @@
...
+
+Building a custom action is an accelerated way of building an action via the HTTP connector, as the custom action utilizes the connector's existing authorization framework and already understands the API for that app.
+
+## App support for custom actions
+Most apps on Workato support custom actions. Custom actions can be found in the actions picklist of the connector.
+
+![Selecting the custom action](/assets/images/developing-connectors/custom-actions/custom-action-selection.png)
+*Selecting the custom action*
+
+The custom action will list the scopes available. Typically, you can only read or write to objects that you have scopes for. We see that we have the following scopes for Slack:
```
eeshansim: Sounds a little odd
"We see that we have the following scopes for Slack:"

Maybe change it to
"For example, when we select custom actions for the Slack connector, we see that only the following scopes are included:"

developing-connectors/custom-actions.md
```diff
@@ -0,0 +1,200 @@
...
+|---------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| Action name                     | Give the custom action you're building a name. This shows up in the action title.                                                                                 |
+| Request type                    | HTTP method of the API endpoint you're calling.                                                                                                                   |
+| Path                            | URL path of the endpoint you're calling. Absolute URI or relative URI are accepted.  Use the absolute URI if you wish to use a different version of the API, etc. |
+| URL params (For GET and DELETE) | If the API endpoint you're calling is a GET or a DELETE, you can pass in URL params.                                                                              |
```
eeshansim: Can we remove "(For GET and DELETE)" from the Input field and mention in description that This input field will appear only if user selects GET or DELETE Request type. Also, we should mention that all values will be URL encoded automatically.

Example:

If the Request type is a GET or DELETE, this input will appear. This input is used to construct URL parameters that will be included in the API request. Values provided here will be URL encoded before being sent.

developing-connectors/custom-actions.md
```diff
@@ -0,0 +1,200 @@
...
+| Input (For POST, PUT and PATCH) | If the API endpoint you're calling is a POST, PUT or a PATCH, you can pass in a JSON request body.                                                                |
```
eeshansim: Can we remove "For POST, PUT and PATCH)" from the Input field and mention in description that this input field will appear only if user selects POST, PUT and PATCH Request type.

Example:

If the Request type is a POST, PUT or a PATCH, this input will appear. This input is used to construct a JSON request body.

We should change the label to Payload instead of Input in the product (quite confusing) WDYT?

developing-connectors/custom-actions.md
```diff
@@ -0,0 +1,200 @@
...
+
+![Naming your custom action](/assets/images/developing-connectors/custom-actions/name-custom-action.gif)
+*Naming your custom action*
+
+Select the HTTP method of the request, based on the API endpoint you're calling. The API documentation should tell you the HTTP method. Slack's API tells us that `list channels` uses a GET method.
+
+![Slack's API documentation for list channels](/assets/images/developing-connectors/custom-actions/slack-list-channels-http-method-docs.png)
+*Slack's API documentation for list channels*
+
+Once selected, relevant fields show up in the action for you to fill.
```
eeshansim: ...relevant **input** fields show up...

developing-connectors/custom-actions.md
```diff
@@ -0,0 +1,200 @@
...
+
+## Custom action input fields
+This section details the configuration input fields in the custom action.
+
+| Input field                     | Description                                                                                                                                                       |
+|---------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| Action name                     | Give the custom action you're building a name. This shows up in the action title.                                                                                 |
+| Request type                    | HTTP method of the API endpoint you're calling.                                                                                                                   |
+| Path                            | URL path of the endpoint you're calling. Absolute URI or relative URI are accepted.  Use the absolute URI if you wish to use a different version of the API, etc. |
```
eeshansim: We should mention the behaviour of using relative path clearly. It can be a very difficult one to troubleshoot if used wrongly. Essentially, if the user adds `/` in front of the value in the input field, the path in the base URI will be overridden. Consider adding this to the description.

> Relative URI will be appended to the end of the base URI. Do not use `/` in your relative URL unless you wish to override the default path of the base URI.
> 
> For example, the base URI of the Slack connector is `https://slack.com/api/`. If you use the value `channel.list` here, the request will be sent to `https://slack.com/api/channels.list`.
> 
> However, if you use the value `/channel.list`, the request will be sent to `https://slack.com/channels.list`.
> 
> Using `/` in front of a relative URI value is useful when you wish to work with a different base path. (like an older API version)

developing-connectors/custom-actions.md
```diff
@@ -0,0 +1,200 @@
...
+*Completed Slack custom action to list channels*
+
+When the above action was tested, we can see the input passed into the action in the detailed job history.
+
+![Job history - input data](/assets/images/developing-connectors/custom-actions/list-channels-action-input.png)
+*Job history - input data*
+
+We can also see the detailed output returned by the Slack API and review if the API call was successful. As `ok` is `true`, it looks like this call was successful.
+
+![Job history - output showing channel data returned](/assets/images/developing-connectors/custom-actions/list-channels-action-input.png)
```
eeshansim: I think there's a mistake here. Should be using output image instead of input.

developing-connectors/custom-actions.md
```diff
@@ -0,0 +1,200 @@
...
+
+![Filling in values for params](/assets/images/developing-connectors/custom-actions/fill-in-param-values.gif)
+*Filling in values for params*
+
+There's just one last thing to provide - the output schema. This can be typically copied from the API documentation and pasted into Workato. Here's the JSON response in Slack.
+
+![Slack API documentation for list channels response](/assets/images/developing-connectors/custom-actions/slack-docs-list-channels-api-response.png)
+*Slack API documentation for list channels response*
+
+```
```
eeshansim: A bit bulky here because of the length. Maybe we can use a table with 2 columns to present them side by side? (helps to reduce the length of this document)

developing-connectors/custom-actions.md
huilinn: @eeshansim, I think it's easier for the user to distinguish if we declare that in the input form column.

Agreed about your other comments, I'll change them for both URL params and input

developing-connectors/custom-actions.md
huilinn: @eeshansim would this only be JSON body?

I think it's called input and output in the product currently, which makes sense as a pair.

developing-connectors/custom-actions.md
huilinn: Removed the image

PR #512, opened 2018-06-11
Workbot update message doc

jandonyada:


workbot/workbot-update-message.md
```diff
@@ -0,0 +1,58 @@
...
+date: 2018-05-21 10:23:00 Z
+---
+
+# Updating Workbot messages
+
+![Update message example](/assets/images/workbot/workbot-update-message/update-message-example.gif)
+*Message updating from button, menu, and back to button again*
+
+You can update a bot message that was posted earlier by using a post message action. This is especially useful when you want to:
+- remove buttons or access menus after users have made their selection (to prevent spam), replacing them with meaningful responses
```
eeshansim: This is a little disorienting. Can we use a table here to organise the use cases and gif?

workbot/workbot-update-message.md
```diff
@@ -0,0 +1,58 @@
...
+
+  ![Reduce clutter](/assets/images/workbot/workbot-update-message/reduce-clutter.gif)
+
+## How it works
+
+You can instruct Workbot to use the message contents of a **Post message** action to update an existing bot message.
+
+Each message posted in Slack has a <kbd>Message ID</kbd>, displayed in the output of Workbot triggers & actions.
+
+  >For Workbot triggers, <kbd>Message ID</kbd> only exists when the trigger is invoked by a user interactions in a Workbot message (e.g. button / menu option clicks or dialog submissions). In such a case, the <kbd>Message ID</kbd> would correspond to the bot message containing the interactive component(s).
```
eeshansim: Is this a quote from another doc? Maybe include the link for users to have some context.

workbot/workbot-update-message.md
```diff
@@ -0,0 +1,58 @@
...
+
+Hence, the <kbd>Message ID</kbd> datapill can be obtained from the output of a **Post command** trigger or a **Post message**  / **Post command reply** action.
+
+## Using 'Message to update'
+
+To update a message, use its <kbd>Message ID</kbd> in the **Message to update** field of a new **Post message** action. This field can be found under the 'Advanced' section of the action.
+
+![Message to update example](/assets/images/workbot/workbot-actions/message-to-update-example.png)
+
+The message will be replaced by the new message defined in the **Post message** action, along with any interactive components like buttons or message menus.
```
eeshansim: "The original message will be replace by the new message..."

Added original to "improve" clarify.

workbot/workbot-update-message.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Replacing messages containing buttons & message menus
+To replace messages containing buttons, message menus, or both, simply add an 'unmapped' button or message menu (depending on what you're replacing) in the updating post message action.
```
eeshansim: Does this mean you cannot replace them with new buttons?

workbot/workbot-update-message.md
```diff
@@ -0,0 +1,58 @@
...
+  To ensure that you're updating the correct message, make sure that you use the <kbd>Message ID</kbd> from the output step which posted the original message.
+
+
+2. **Can't find the 'Message to update' field**
+  Make sure that it's checked from the 'Add/remove optional fields' list at the bottom of the post message action. After it's enabled, the field can be found in the Advanced group of the post message action.
+  ![Message to update optional fields](/assets/images/workbot/workbot-update-message/message-to-update-optional-fields.gif)
+
+
+3. <kbd>Message ID</kbd> **pill is empty**
+  When using a <kbd>Message ID</kbd> pill from the output of a trigger, make sure that the trigger is invoked by a button / menu option click or dialog submission from a bot message (presumably from another recipe). Only triggers that are invoked from interactions with a posted bot message will have a <kbd>Message ID</kbd>.
```
eeshansim: Not sure I understand this paragraph. Are you saying that commands invoked by manually typing the command will not have a message ID? We may need to include best practice for how a user can enforce this.

workbot/workbot-update-message.md
jandonyada: Hmm ok I can change it to table form.

workbot/workbot-update-message.md
jandonyada: Not quite... Wanted to highlight the reason why a trigger can also have a Message ID - thought that a quote would illustrate it better.

Because Callback IDs point to a message, command triggers normally won't have one. But if triggered by a bot command (like button / menu option clicks or dialog submit), then the trigger's output _will_ have the Callback ID of the message containing that bot command component (i.e. the message that posted the button, menu option or dialog).

workbot/workbot-update-message.md
jandonyada: Yea, when a user invokes a command, the trigger output will not have a message ID because there is no message to update yet.

eeshansim: Ok, let's state that to make it clear.

Any other way to illustrate that?

workbot/workbot-update-message.md
jandonyada: You can - any button(s) posted by the updating message will replace the buttons in the original message. Hmm I guess it could be conveyed clearer.. I'll think about it. Any advice?

jandonyada: But as far as I know, there's a clearer use case to remove buttons (i.e. anti-spam) than there is to replace a button with a new one.

workbot/workbot-update-message.md
jandonyada: Would it be clearer to explain this way:
>Because Callback IDs point to a message, command triggers normally won't have one. But if triggered by a bot command (like button / menu option clicks or dialog submit), then the trigger's output will have the Callback ID of the message containing that bot command component (i.e. the message that posted the button, menu option or dialog).

Was trying to avoid a wall of text...

eeshansim: Second one is better. Can we keep the phrasing consistent with "invoke".

"When a user invokes this trigger by typing a command" vs "When a user invokes a command"

First one fits better with "if the trigger is invoked from a bot command" IMO

workbot/workbot-update-message.md
eeshansim: Yep ok we should be use case oriented. This is fine.

workbot/workbot-update-message.md
eeshansim: ok, same point as below I guess.

workbot/workbot-update-message.md
jandonyada: Good point, changing it to the prior.

workbot/workbot-update-message.md
eeshansim: you missed one here

jandonyada: Updated.

jandonyada:
@eeshansim updated

PR #513, opened 2018-06-12
Error handling best practices article

graceliching:
Covers how recipe building should be done to account for errors


SUMMARY.md
```diff
@@ -32,6 +32,7 @@
     * [Preview data](recipes/data-pills-and-mapping.md#preview-data)
     * [Fields mapping](recipes/data-pills-and-mapping.md#fields-mapping)
     * [Data transformation](recipes/data-pills-and-mapping.md#data-transformation)
+  * [Error handling best practices](best-practices-error-handling.md)
```
eeshansim: wrong path. Should be `recipes/best-practices-error-handling.md`.

recipes/best-practices-error-handling.md
```diff
@@ -0,0 +1,46 @@
...
+
+# Error handling and monitoring best practices
+
+This article covers the best practices for recipe building for handling errors. View this [recipe](https://www.workato.com/recipes/696621) that showcases all the best practices.
+
+![Error Handling Recipe](/assets/images/building-best-practices/error-handling-eg1.png)
+
+**1. Error monitor step**
+
+Always wrap all the actions within your recipe within an error monitoring step so that all errors, user created or system generated can be universally caught and handled. To do this, click on 'Add a new step', then on 'Action with error monitor'.
```
eeshansim: "Always wrap all the actions within your recipe..."

Must all actions use this? When is it not needed? Not very comfortable with saying this because users will be led to think all recipes MUST have this.

By default, all recipe errors will result in an email notification with relevant information. Monitor with error helps with continuing a job despite an error and retrying actions.

recipes/best-practices-error-handling.md
```diff
@@ -0,0 +1,46 @@
...
+
+![error monitor step2](/assets/images/building-best-practices/error-handling-error-monitor2.png)
+
+The recipe will perform the steps nested in 'On error'. In the example, the recipe will send out an email via Workato to the administrator's email address when the actions have been retried three times and failed. 
+
+![error monitor step3](/assets/images/building-best-practices/error-handling-eg2.png)
+
+2. **Conditional actions**
+
+Try as much as possible to prevent errors by checking the input values to key actions in your recipe, i.e. when you are writing to a destination system. Using conditional actions and stop actions to validate the incoming data fails the recipe early to reduce unnecessary processing and API calls. This will speed up the rate of running your jobs. You can add conditional actions by clicking on 'Add a new step', then on 'Conditional Action'.
```
eeshansim: can we say destination **application** instead of **system**

eeshansim: We should also mention the classic example of conditional actions and what it does for the user

1. Search using unique values
2. Conditional action for presence of existing record
   - Update existing record
3. Conditional action for absence of existing record
   - Create new record

What this does? Prevent duplicate records or errors resulting from creation of duplicate records.

WDYT? Should we include this here?

eeshansim: Rephrase:
"Use conditional actions to validate the incoming data and stop actions to end or fail a recipe job early to reduce unnecessary processing time and API calls. This will speed up the time taken for all your jobs."

recipes/best-practices-error-handling.md
```diff
@@ -0,0 +1,46 @@
...
+date: 2018-06-11 10:00:00 Z
+---
+
+# Error handling and monitoring best practices
+
+This article covers the best practices for recipe building for handling errors. View this [recipe](https://www.workato.com/recipes/696621) that showcases all the best practices.
+
+![Error Handling Recipe](/assets/images/building-best-practices/error-handling-eg1.png)
+
+**1. Error monitor step**
```
eeshansim: We don't usually number our headers

recipes/best-practices-error-handling.md
graceliching: I think we can discuss building up this article to make it wholesome as a 'best practices' article or include that as a separate article. This one is focused on how to handle errors more effectively. We have an old version in the repo (not published), but I feel that one isn't very helpful. 

In the case of the classic example of conditional actions, i think it should be included in the conditional actions section or in another best practices article.

eeshansim: ok. What about the rephrase? It reads a bit funny now.

recipes/best-practices-error-handling.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Error handling and monitoring 
+
+This section covers the best practices for recipe building for handling errors. View this [recipe](https://www.workato.com/recipes/696621) that showcases all the best practices.
+
+![Error Handling Recipe](/assets/images/building-best-practices/error-handling-eg1.png)
+
+1. Error monitor step
+
+To monitor specific steps for errors, wrap all the actions within your recipe within an error monitoring step. This allows Workato to monitor all errors, whether user created or system generated, so that they can be universally caught and handled. To do this, click on 'Add a new step', then on 'Action with error monitor'.
```
eeshansim: Need to include that any errors within the monitor step will not cause the job to stop.

graceliching:
@eeshansim addressed your comment, can you review and merge? thanks!

PR #533, opened 2018-07-03
Updated permissions

jandonyada:


connectors/jira.md
```diff
@@ -42,74 +42,149 @@ Password to connect to Jira.
...
+For example, in the project permission scheme below, only the 'QA' role has been permitted to perform the 'Close Issues' action.
+![QA role](/assets/images/jira-docs/qa-role.png)
 
-![Permission schemes](/assets/images/jira-docs/roles-permission-schemes.PNG)
+Hence, if the Workato recipe wants to perform any 'Close issue' actions, the linked Jira account must also be assigned the role 'QA' for the action to be authorized.
 
-##### 2. Applications
-For permissions across **Applications**, **Any logged in user** permission must be granted to use the Workato triggers and actions.
+##### 2. Application access
+Application access settings let you control which people have access to which products. If your site only has one product (i.e. if you have a Confluence-only or Jira-only instance, for example) – users are automatically granted access to that product when they sign up.
```
eeshansim: is the term "people" specific to Jira? If not, it's better to use the word person.

"Application access settings let you control which ~people~ person have access to..."

jandonyada: True, updating as per your suggestion.

PR #541, opened 2018-07-20
Jira actions doc

jandonyada:


connectors/jira/actions.md
```diff
@@ -0,0 +1,1465 @@
+---
+title: Jira actions
+date: 2018-07-18 06:15:00 Z
+---
+
+# Jira actions
+The following Jira actions can be used to execute workflows in your recipes.
+
+### Requirements
+The linked Jira account must have sufficient permissions to the respective resource in each action. For more information on Jira permissions, head over to our [Jira project permissions documentation](https://docs.workato.com/connectors/jira.html#roles-and-permissions-required-to-connect).
```
huilinn: Learn more with our [Jira project permissions documentation](https://docs.workato.com/connectors/jira.html#roles-and-permissions-required-to-connect).

connectors/jira/actions.md
```diff
@@ -0,0 +1,1465 @@
...
+    <tr>
+      <td>Custom fields</td>
+      <td>Miscellaneous fields</td>
+      <td>Custom fields defined by your Jira field configurations. </td>
+    </tr>
+  </tbody>
+</table>
+
+#### Project issue type
+Selecting a project issue type will retrieve the issue's fields (as specified by your Jira issue type scheme) for you to populate with datapills. You can select a project issue type from the picklist provided. This also retrieves the corresponding issue type's fields.
```
huilinn: Sounds a bit repetitive. Maybe:
"Selecting a project issue type will retrieve the corresponding issue type's fields (as specified by your Jira issue type scheme) for you to populate with datapills."

connectors/jira/actions.md
```diff
@@ -0,0 +1,1465 @@
...
+
+![Project issue type picklist](/assets/images/jira-docs/project-issue-type-picklist.gif)
+*Specifying a project issue type using the picklist*
+
+You can also use the project **Key** and issue type **Name** datapills, separated by two hyphens '--'e.g. <kbd>Key</kbd>--<kbd>Name</kbd> to dynamically generate the project issue type. This is especially useful if you want to dynamically create issues across a range of projects and issue types.
+
+![Project issue type dynamic](/assets/images/jira-docs/project-issue-type-dynamic.gif)
```
huilinn: This gif seems very different from the others in style, and the lines and datapills don't show up properly. Can we fix that?

connectors/jira/actions.md
```diff
@@ -0,0 +1,1465 @@
...
+
+Alternatively, you can also hardcode the project & issue type by manually entering the project key and the issue type, again separated by two hyphens, e.g. **PROJ--Bug**.
```
huilinn: Shall we move this to line 70 (right after the picklist option) as it's equivalent to the picklist?

Alternatively, we don't really have to cover this as picklist is much easier.

connectors/jira/actions.md
jandonyada: Hmm not sure what I should change here - what's the difference in style?

As for lines and datapills now showing properly, can you share what you're looking at? Looks ok on my end:
project-issue-type-picklist

connectors/jira/actions.md
jandonyada: That's true, I will remove it.

connectors/jira/actions.md
huilinn: The dynamic one (not the picklist one) looks like this for me
[Image from Gyazo](https://gyazo.com/902b052e09613953642d3897b648deab)

connectors/jira/actions.md
```diff
@@ -0,0 +1,1460 @@
...
+      <td>Miscellaneous fields</td>
+      <td>Custom fields defined by your Jira field configurations. </td>
+    </tr>
+  </tbody>
+</table>
+
+#### Project issue type
+Selecting a project issue type will retrieve the corresponding issue type's fields (as specified by your Jira issue type scheme) for you to populate with datapills
+
+![Project issue type picklist](/assets/images/jira-docs/project-issue-type-picklist.gif)
```
huilinn: Going forward, we should try to use real examples as much as possible

connectors/jira/actions.md
```diff
@@ -0,0 +1,1460 @@
...
+*Specifying a project issue type using the picklist*
+
+You can also use the project **Key** and issue type **Name** datapills, separated by two hyphens '--'e.g. <kbd>Key</kbd>--<kbd>Name</kbd> to dynamically generate the project issue type. This is especially useful if you want to dynamically create issues across a range of projects and issue types.
+
+![Project issue type dynamic](/assets/images/jira-docs/project-issue-type-dynamic.gif)
+
```
huilinn: Maybe add this to let user know they should use dynamic entry with sample project issue type field?

"If you are creating or updating issues dynamically, via project **Key** and issue type **Name** datapills, Workato will not be able to generate your issue input fields at design-time. This is because we don't know what fields to fetch from Jira yet. In such cases, you should use the [sample project issue type](/connectors/jira/actions.md#sample-project-issue-type) to generate input fields.

connectors/jira/actions.md
```diff
@@ -0,0 +1,1460 @@
...
+Because issues can have different fields in different projects, you may have a sample project issue type (defined in Jira) that contains all possible issue fields across multiple projects & issue types.
+
+Specifying the sample project issue type here retrieves all fields related to it, and displays them in your recipe. This allows you to map datapills to the new issue(s) being created.
+
+This way, you can perform the mapping of datapills that works for all issues dynamically created across multiple projects & issue types.
+
+![Sample project issue type](/assets/images/jira-docs/sample-project-issue-type.png)
+*Using a sample project issue type to retrieve fields across multiple issue types & projects*
+
+Similar to the [project issue type field](/connectors/jira/actions.md#project-issue-type) above, you can use the picklist to choose the project & issue type, or enter them manually by entering the project key & issue type, separated by two hyphens, e.g. **PROJ--Bug**.
```
huilinn: Let's not highlight the manual part here, there's no point in asking user to put manual entry in. We probably shouldn't have that toggle option in the connector...

connectors/jira/actions.md
```diff
@@ -0,0 +1,1460 @@
...
+Summary of the issue to be created. By default, this field is required, unless defined otherwise by your Jira field configuration.
+
+#### Reporter
+The person reporting this issue. Use the Jira username of the reporter, e.g. if reporter's e-mail address is **johndoe@workato.com**, use **johndoe**.
+
+#### Description
+Description of the issue.
+
+#### Issue priority
+The priority of the issue. This list is retrieved as defined by the defined list of picklist values in Jira. Hence, the values may vary from the screenshot below.
```
huilinn: How about:

"The priority of the issue. Valid values from your Jira account are fetched."

connectors/jira/actions.md
```diff
@@ -0,0 +1,1460 @@
...
+#### Troubleshooting
+
+##### Unable to see project issue type(s)
+If you don't see the project issue type(s) you're looking for, the linked Jira account may not have sufficient permissions to view the project or issue/issue type.
+
+If you are a Jira administrator of your Jira workspace, check the linked Jira account's permissions by using Jira's Permissions helper. See our [Jira permissions documentation](/connectors/jira.html#jira-permissions-helper) or check out [Jira's guide on Permissions helper](https://confluence.atlassian.com/adminjiracloud/using-the-permission-helper-868982879.html) for more details. Otherwise, check with your Jira administrator(s) for help.
+
+##### Unable to see field(s) / custom field(s)
+If you don't see the field(s) you're looking for,
+1. The linked Jira account may not have sufficient permissions to view the field(s). Again, you can use Jira's Permissions helper if you're a Jira administrator. Otherwise, check with your Jira administrator(s) for help.<br><br>
```
huilinn: Instead of saying "Again..." and not providing the link, each section should be discrete, e.g. user who's reading through one section might not have gone through all sections organically. Let's provide a link here as well to permissions helper and our jira permissions documentation?

connectors/jira/actions.md
jandonyada: Ah, this is much better, thanks!

connectors/jira/actions.md
jandonyada: What about "Valid picklist values from your Jira account are fetched."?

connectors/jira/actions.md
jandonyada: Good point, adding the links.

PR #545, opened 2018-07-31
Quick Base csv bulk action doc. For Learn more link

zweedao:
Try out the new action here: https://preview.workato.com/recipes/25445#recipe


zweedao:
@graceliching can you help to review this soon :) We want to release this action this week.

connectors/quick-base.md
```diff
@@ -0,0 +1,9 @@
+---
```
graceliching: @zweedao you need to change the SUMMARY file as well to include these 2 docs.

connectors/quick-base/action-import-csv.md
```diff
@@ -0,0 +1,107 @@
+---
+title: Workato connectors - Quick Base's action - Create and update records in bulk from CSV file
+date: 2018-07-30 06:00:00 Z
+---
+
+#Quick Base's action: Create and update records in bulk from CSV file
```
graceliching: missing space after #

graceliching: since this is under the quick base connector doc, you don't need "Quick Base's action". You can name it "Create/update records in bulk via CSV file" to keep it consistent with Salesforce's article.

connectors/quick-base/action-import-csv.md
```diff
@@ -0,0 +1,107 @@
...
+![Quick Base's action: Create and update records in bulk from CSV file](/assets/images/connectors/quick-base/action-import-csv.png)
+
+## 1. How it works
```
graceliching: Remove the numbers from the headers, this is not a convention we use

connectors/quick-base/action-import-csv.md
```diff
@@ -0,0 +1,107 @@
...
+This action takes a CSV file then create or update Quick Base records in bulk. Note that this action is not an Upsert operation, the reason of which we will explain below.
+
+The action description text already gives you an overview of how it works. In this document, we will will discuss a few important points when using this action:
```
graceliching: Users will not know what action description text you are referring to. I think it's best to remove this statement as it does not tell users any useful info.

connectors/quick-base/action-import-csv.md
```diff
@@ -0,0 +1,107 @@
...
+
+### 1.1 Create records vs. Create & Update records
+This action can either create new Quick Base records only, or create & update Quick Base records at the same time. The way to control this behaviour depends on your provided CSV file and field mappings.
+
+####Create new records only
```
graceliching: missing space

connectors/quick-base/action-import-csv.md
```diff
@@ -0,0 +1,107 @@
+---
+title: Workato connectors - Quick Base's action - Create and update records in bulk from CSV file
+date: 2018-07-30 06:00:00 Z
+---
+
+#Quick Base's action: Create and update records in bulk from CSV file
+![Quick Base's action: Create and update records in bulk from CSV file](/assets/images/connectors/quick-base/action-import-csv.png)
+
+## 1. How it works
+This action takes a CSV file then create or update Quick Base records in bulk. Note that this action is not an Upsert operation, the reason of which we will explain below.
```
graceliching: Can we move this statement into the create/update records section? It doesn't really make sense on its own. Let's write docs with the assumption that users are not likely to read through the whole article, but go to the section they are working on in recipes

connectors/quick-base/action-import-csv.md
```diff
@@ -0,0 +1,107 @@
...
+
+The action description text already gives you an overview of how it works. In this document, we will will discuss a few important points when using this action:
+
+### 1.1 Create records vs. Create & Update records
+This action can either create new Quick Base records only, or create & update Quick Base records at the same time. The way to control this behaviour depends on your provided CSV file and field mappings.
+
+####Create new records only
+Workato uses Quick Base record ID to update records. If you only want to create new Quick Base records, just leave blank the first field in the `Column mappings` section, which is your table's `Record ID#` (Or if you use custom record ID column in your table, such as `Order ID` or `Customer ID`, this field name will change accordingly).
```
graceliching: Workato searches for matches in the Quick Base record ID field to update records. If you only want to create new Quick Base records, just leave the first field in the `Column mappings` section, which is your table's `Record ID#`, blank. If you use custom record ID column in your table, such as `Order ID` or `Customer ID`, this field name will change accordingly.

connectors/quick-base/action-import-csv.md
```diff
@@ -0,0 +1,107 @@
...
+
+![Blank Record ID](/assets/images/connectors/quick-base/csv-import-record-id-blank.png)
+####Create & update records at the same time
```
graceliching: missing space

connectors/quick-base/action-import-csv.md
```diff
@@ -0,0 +1,107 @@
...
+
+When some rows failed, however, the action may still be considered "complete". It is because Quick Base has successfully accepted the CSV file, but some rows failed to be created/updated into Quick Base records. Quick Base API will send back a "success" response with list of successful and failed records.
+
+In this example, the job report shows "Complete" status. However, when we check the action output, there is actually 1 failed chunk.
+![Job complete](/assets/images/connectors/quick-base/csv-import-job-complete.png)
+
+![Chunk error](/assets/images/connectors/quick-base/csv-import-chunk-error.png)
+
+It is thus important to always handle failed CSV rows in your recipe. The output pill `CSV contents of failed records` contains all failed CSV rows. You can use this to save the failed rows into a CSV file. Then check the job report for error reason, fix those failed rows and re-import them later. Here is [a sample recipe](https://preview.workato.com/recipes/25445#recipe) with failed-rows handling:
```
graceliching: 1. Do you have to save the failed rows into a CSV file? What if they do not use box and there's no action that allows this on another adapter? If they do not import the failed rows into Box, can you provide some alternatives? eg. write to gsheets or download to their computer as a csv file.

2. If you are doing bulk create, how do users re-import if there is no record ID? Will rerunning the job create duplicates? If so, we should include a section on how they should search for existing records first.

connectors/quick-base/action-import-csv.md
```diff
@@ -0,0 +1,107 @@
...
+
+![Handling failed CSV rows](/assets/images/connectors/quick-base/csv-import-error-handling.png)
+
+## 2. Input fields
```
graceliching: Configuring input fields

connectors/quick-base.md
zweedao: not changing SUMMARY file for now yet, so users will not access the docs before the action is in production.

connectors/quick-base/action-import-csv.md
zweedao: actually if you look at all database docs, they are in this format. Each trigger/action is in a separate page, cleaner & shorter, so adding app name make it clearer. 

While for Salesforce, all trigger/actions are on 1 page so no need app name, but that style makes the page long & messy I think.

connectors/quick-base/action-import-csv.md
zweedao: Our doc's H1, H2, H3 styling makes it difficult to distinguish headers & sections. Reading feels a bit lost sometimes. Can we either apply a numbering convention for header or improve the header styling (will need engineers)?

connectors/quick-base/action-import-csv.md
zweedao: 1. Don't have to but recommended, that's the easiest way. 
All file connectors have ability to upload file. 
Other alternatives like GSheets requires more complex recipe, and then needs to export into a CSV to use with this action anyway.  Will add a sentence suggesting some other file connectors.

2. The `CSV contents of failed records` has Record ID. No duplicate when importing again, because those are all failed rows that were not created/updated yet.

connectors/quick-base/action-import-csv.md
graceliching: # Quick Base action: Create/update records in bulk via CSV file

connectors/quick-base/action-import-csv.md
graceliching: For 1, not all users will use file connectors. 

For 2, I think you need another paragraph explaining what to do after saving the file. it is not explicit.

connectors/quick-base/action-import-csv.md
zweedao: will remove

connectors/quick-base/action-import-csv.md
graceliching: Currently, let's leave it as is. We don't have number styling anywhere else in the docs and I don't really see how it improves readability in this case. Technically, the section you number '2' (input fields) still refer to the same action, which is in one section on its own.

connectors/quick-base/action-import-csv.md
```diff
@@ -29,19 +29,21 @@ Then in the `Column mappings` section, map your table's `Record ID#` column with
...
+- If record ID is present, search for the matching record ID in Quick Base then update that record. If no matching record ID is found in Quick Base, that CSV row will be fail.
 
 ### 1.2 Handling failed CSV rows:
 This action uses [batch processing](https://docs.workato.com/features/batch-processing.html), so it will divide your CSV file into smaller chunks of rows (or batches) then submit to Quick Base. When 1 row in a chunk failed to be created/updated into Quick Base records, Quick Base will reject that whole chunk, but other chunks will not be affected.
 
-When some rows failed, however, the action may still be considered "complete". It is because Quick Base has successfully accepted the CSV file, but some rows failed to be created/updated into Quick Base records. Quick Base API will send back a "success" response with list of successful and failed records.
+When some rows failed, however, the action may still be considered "complete". It is because Quick Base has successfully accepted the CSV file. Quick Base API will then send back a "success" response with list of successful and failed records.
```
graceliching: When some rows fail

connectors/quick-base/action-import-csv.md
```diff
@@ -0,0 +1,109 @@
...
+![Mapped Record ID](/assets/images/connectors/quick-base/csv-import-record-id-mapped.png)
+
+Workato uses these record ID columns to decide whether to create or update records. The rule is that for each CSV row:
+- If record ID is empty, create a new Quick Base record.
+- If record ID is present, search for the matching record ID in Quick Base then update that record. If no matching record ID is found in Quick Base, that CSV row will be fail.
+
+### 1.2 Handling failed CSV rows:
+This action uses [batch processing](https://docs.workato.com/features/batch-processing.html), so it will divide your CSV file into smaller chunks of rows (or batches) then submit to Quick Base. When 1 row in a chunk failed to be created/updated into Quick Base records, Quick Base will reject that whole chunk, but other chunks will not be affected.
+
+When some rows failed, however, the action may still be considered "complete". It is because Quick Base has successfully accepted the CSV file. Quick Base API will then send back a "success" response with list of successful and failed records.
```
graceliching: when some rows fail

connectors/quick-base/action-import-csv.md
```diff
@@ -0,0 +1,109 @@
+---
+title: Workato connectors - Quick Base's action - Create and update records in bulk from CSV file
+date: 2018-07-30 06:00:00 Z
+---
+
+# Quick Base's action: Create and update records in bulk from CSV file
+![Quick Base's action: Create and update records in bulk from CSV file](/assets/images/connectors/quick-base/action-import-csv.png)
+
+## 1. How it works
```
graceliching: Can we remove the numbers for now? I don't see how this numbering improves readability in this case. For example, the section that is numbered '2' (input fields) is still related to configuring the same action, which is numbered '1'. This action is already it's own section

zweedao: Actually I try to organize this way:
- Section 1 is about How it works & important things to note.
- Section 2 is about Input.
- Section 3 is about Output.

Removed the numbers for now. My eyes hurt 😭 , now I can't distinguish which section is which. All the headers look so similar in size & color. It's like an infinite document. Not easy for user to look for what they want. Long document like Salesforce is even worse.

Think you really should come up with the improved header styling.

connectors/quick-base/action-import-csv.md
zweedao: I think let's keep the `action` and exact action name. So it's clear it's about this action, and not look like a tip-and-trick / how-to article.

zweedao:
@graceliching ok addressed all points alrd. Can you take a look?

zweedao:
@eeshansim can help to merge this?

PR #563, opened 2018-08-09
552 db permissions

eeshansim:


connectors/redshift.md
```diff
@@ -44,6 +44,52 @@ The Redshift connector uses basic authentication to authenticate with Redshift.
...
+
+```sql
+GRANT SELECT,INSERT ON TABLE customer TO workato;
+```
+
+Finally, check that this user has the necessary permissions. Run a query to see all grants.
+
+```sql
+SELECT
+    u.usename,
```
jandonyada: I see a lot of `usename` here. Should it be `username`?

eeshansim: `usename` is the actual column name for Redshift.

PR #575, opened 2018-08-20
add docs for sql select

### LINKED ISSUE #574: PostgreSQL custom SQL select action

eeshansim:
fixes #574

### DIFF
--- a/connectors/postgresql/select.md
+++ b/connectors/postgresql/select.md
@@ -43,3 +43,19 @@ This input field determines the maximum number of rows to return. The default li
 
 ### Offset
 This input field gives you the option to fetch only a page of results from the entire results set. For example, to skip the first 100 rows of the selected results set, input `100` to this field. The default is `0`.
+
+## Select rows using custom SQL
+This action lets you select rows based on a custom SQL query. Rows that are returned from the query will be returned as the output of this action.
+
+![Select rows using custom SQL action](/assets/images/oracle/custom-sql-action.png)
+*Select rows using custom SQL action*
+
+### SQL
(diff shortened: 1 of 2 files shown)


connectors/postgresql/select.md
```diff
@@ -43,3 +43,19 @@ This input field determines the maximum number of rows to return. The default li
...
+## Select rows using custom SQL
+This action lets you select rows based on a custom SQL query. Rows that are returned from the query will be returned as the output of this action.
+
+![Select rows using custom SQL action](/assets/images/oracle/custom-sql-action.png)
+*Select rows using custom SQL action*
+
+### SQL
+Provide the SQL query to be executed to select rows. The query here will be used to generate the output datatree. To do this, the SQL will be executed once when you provide it. You can map datapills here to execute dynamically changing SQL statements. Remember to wrap datapills in quotes (`''`).
+
+Avoid using limit clause like `LIMIT` in your SQL. This is because the limit to the number of rows returned in the query is based on the value defined in the [**Limit** input field](#limit-1). Adding your own limit clause will cause the action to fail.
```
jandonyada: Avoid using limit clause**s** like....

I'm assuming the SQL query can have multiple instances of `LIMIT` here.

jandonyada: The **Limit** input field is described again below - should the anchor be (#limit-2) instead? Otherwise, since it's just below you may consider not using the anchor at all.

eeshansim: The first Limit uses the anchor `#limit`, second one uses `#limit-1`.

jandonyada: I see, my bad. In that case it's all good.

PR #585, opened 2018-09-03
Product updates 3 sept

graceliching:
Some links yet to be added as docs not merged/posts not published.

### DIFF
--- a/product-updates.md
+++ b/product-updates.md
@@ -14,31 +14,31 @@ Our engineers are hard at work daily to make Workato better! View the latest upd
 
 ### Platform enhancements
 
-#### Stronger security with TLS 1.1 update
+- **Stronger security with TLS 1.1 update**
 
-**What's new:** Workato is [upgrading to TLS 1.1 or higher](https://product.workato.com/2018/08/07/stronger-security-with-tls-1-1-update/) in order to align with industry best practices for security and data integrity.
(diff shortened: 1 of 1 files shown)


product-updates.md
```diff
@@ -8,6 +8,49 @@ date: 2018-06-05 21:45:00 Z
 
 Our engineers are hard at work daily to make Workato better! View the latest updates to our platform here.
```
jandonyada: When I read this, I thought there was a missing link due to how often we use 'here' as a link (which we should try to do less of, IMO).
Not part of the commit, but this could be less ambiguous if it said "View the latest updates to our platform on this page." What do you think?

product-updates.md
```diff
@@ -8,6 +8,49 @@ date: 2018-06-05 21:45:00 Z
...
+  
+- **Job report: Timing tab**
+  
+  **What's new:** The timing tab for every step in a recipe has been exposed. Users can find it next to the 'input' and 'output' tabs in the individual job reports. This tab shows the time taken for every action and can be used in debugging recipes. 
+
+### Connector enhancements
+
+- **New connector: On-Prem Command-line Scripts**
+
+  **What's new:** Workato's `On-prem command-line scripts` connector allows you to run whitelisted command line scripts in your private network with an on-premises agent. Find out how to set a profile up [here](https://docs.workato.com/on-prem/profile.html#command-line-scripts-profile).
```
jandonyada: Best to state the location as opposed to 'here', which may be ambiguous. How about:

"Find out how to set up a profile over at our [on-prem profiles documentation](https://docs.workato.com/on-prem/profile.html#command-line-scripts-profile)." ?

PR #606, opened 2018-09-27
Product notifications official document

graceliching:


product-notifications.md
```diff
@@ -0,0 +1,82 @@
...
+
+To communicate product-level information to our users, we make them available on three main sites:
+
+- [Status page](https://status.workato.com/#)
+- [Product updates page](/product-updates.md)
+- [Product blog](https://product.workato.com/)
+
+We recommend that users subscribe to these pages to receive timely updates on the changes and enhancements that happen with our product. Read on for details on the information available on these pages.
+
+## Website and server status
```
jandonyada: My understanding is that this section here is to highlight to the user why they should subscribe to the status page, and what sort of info it contains. Generally, I'd recommend a structure like this:
- what info the page offers
- details about the info
- why they should subscribe

For example,

## Status page
The **[Status page](https://status.workato.com/#)** posts important information related to website & server updates.

The following are classified as website/server updates:
- Website uptime & downtime
- Status of recipe job processing
- Status of on-premise connectivity
- Status of webhook ingestion
- Past incidents that affect recipe performance
 
We recommend that users subscribe to the status page to receive updates. You can do so by clicking on the 'Subscribe to updates' button on the top right of the page.

product-notifications.md
```diff
@@ -0,0 +1,82 @@
...
+
+- Website uptime & downtime
+- Status of recipe job processing
+- Status of on-premise connectivity
+- Status of webhook ingestion
+- Past incidents that affect recipe performance
+
+All information related to these issues are posted on **[status.workato.com](https://status.workato.com/#)**.
+
+We recommend that users subscribe to the status page to receive updates. You can do so by clicking on the 'Subscribe to updates' button on the top right of the page.
```
jandonyada: Let's keep the POV consistent here. Since we're already addressing 'users' earlier in the article, we should say "They can do so by clicking..." instead of "You can do so by clicking...".

Alternatively, we can change the POV to 2nd person and swap all use of 'users' to 'you' — but I'd stick with the former instead.

product-notifications.md
```diff
@@ -0,0 +1,82 @@
...
+
+- Potential recipe-breaking change
+- Changes to security/legal terms
+- Actions/applications that will be removed/deprecated and no longer grandfathered
+- Changes that require you, the user to make edits to their recipes to make sure they continue running
+- Planned downtime (maintenance etc.)
+
+These changes may affect recipe performance and are thus categorised as major changes. At Workato, we try to keep these changes to a minimum to ensure that your recipes do not need to be edited unless absolutely necessary.
+
+Information about these changes will be sent to your Workato email at least **30 days** in advance. Reminder emails will then be sent to you **14 days** and **7 days** prior to the change. One final email will be sent to you once the change is live on our site. The emails will be sent to the addresses provided in the [error notifications settings tab](https://www.workato.com/users/current/edit#notifications).
```
jandonyada: Same thing about POV here

product-notifications.md
```diff
@@ -0,0 +1,82 @@
...
+
+The following are classified as product enhancements/fixes:
+
+- Connector enhancements
+- Bug fixes
+- New platform features
+- New connectors
+- UI changes
+
+If any of these changes were brought to our attention by you through any of our support channels, rest assured that the customer success team will reach out to you directly to inform you of the changes.
```
jandonyada: Same thing about POV here.

Also, "...rest assured that ~the~ **our** customer success team..."

product-notifications.md
```diff
@@ -0,0 +1,82 @@
...
+
+These changes are completely backward compatible will not require you to take any action. All this information can be found on the **[Product updates page](/product-updates.md)**. Product-level highlights and monthly updates will also be written in more detail on the **[Product blog](https://product.workato.com/)**.
```
jandonyada: Same thing about POV here.

Also, "These changes are completely backward compatible, **and** will not required...."

graceliching:
Thanks @jandonyada I think your comments made sense :) I've edited it acc to them!

PR #607, opened 2018-09-28
Update Quick Base bulk behavior

zweedao:
We discovered new behavior of Quick Base bulk API. Need to update the doc to reflect this. Customer alrd uses this in production, so need to push out asap.


connectors/quick-base/action-import-csv.md
```diff
@@ -7,18 +7,22 @@ date: 2018-07-30 06:00:00 Z
...
 
 ## How it works
-This action takes a CSV file then create or update Quick Base records in bulk. Note that this action is not an Upsert operation: if you provide a record ID that does not exist in Quick Base, this action will reject the record instead of creating a new one.
-
-The followings are key details to understand when using this action:
+This action takes a CSV file then create or update Quick Base records in bulk. Note that this action is not an Upsert operation. The way it works depend on your Quick Base table configuration, which is explained in detail below.
 
 ### Create records vs. Create & Update records
-This action can either create new Quick Base records only, or create & update Quick Base records at the same time. The way to control this behaviour depends on your provided CSV file and field mappings.
+This action can either create new Quick Base records only, or create & update Quick Base records at the same time. The way to control this behaviour depends on your Quick Base table configuration:
```
graceliching: I would edit the structure of this section, it doesn't read well. Let me know what you think of below:

## How it works

This action uses a CSV file to create or update Quick Base records in bulk. Note that this action is not an Upsert operation but either **creates only** or **creates and updates** your Quick Base records.

The way it works depends on your Quick Base table configuration, which is explained in detail below.

### Create records

To only create records (without updating) with your CSV file..... 

#### Table using default Record ID#
Move information here

#### Table uses custom Record ID
Move information here

### Create and Update records
To both create and update your Quick Base table with your CSV file..... 

#### Table using default Record ID#
Move information here

#### Table uses custom Record ID
Move information here

graceliching: This follows the structure of the docs better since it's based on how our actions work vs. how the customer has configured their table

connectors/quick-base/action-import-csv.md
```diff
@@ -27,9 +31,23 @@ Then in the `Column mappings` section, map your table's `Record ID#` column with
...
+
+#### If your Quick Base table uses custom record ID
+![Quick Base Custom Record ID](/assets/images/connectors/quick-base/qb-custom-record-id.png)
+
+Custom record ID, such as Order ID or Customer ID, is not automatically generated by Quick Base. So you always need to provide it in your CSV file.
+
+![Custom ID column in CSV file](/assets/images/connectors/quick-base/csv-custom-id.png)
+
+In Workato, map the first field in `Column mappings` section with the appropriate record ID in your CSV file.
+![Custom Record ID](/assets/images/connectors/quick-base/csv-import-custom-id-mapped.png)
```
graceliching: I would take a .gif or a pic of the entire section since the field could appear twice-- once at the table settings and once at column mappings

zweedao:
@graceliching I re-organized a bit and changed the screenshot. 

At first I also thought to organize by Create only vs Create/Update. But, there is no `Create only` for table using default ID. Also the setup & behaviors of default ID vs custom ID are quite different. So better to separate right away.

connectors/quick-base/action-import-csv.md
```diff
@@ -35,7 +32,7 @@ Based on the provided record IDs, Workato will decide whether to create or updat
 - If record ID is empty, `create` a new Quick Base record.
 - If record ID is present, search for the matching record ID in Quick Base then `update` that record. If no matching record ID is found in Quick Base, that CSV row will `fail`.
 
-#### If your Quick Base table uses custom record ID
+### If your Quick Base table uses custom record ID
 ![Quick Base Custom Record ID](/assets/images/connectors/quick-base/qb-custom-record-id.png)
```
graceliching: [image]

In the example above, the custom field used as a record ID is the `Order ID` field. If the Quick Base table uses a custom record ID, you cannot choose whether the action creates or creates and updates the table. This is because the custom record ID, such as Order ID or Customer ID, is not automatically generated by Quick Base. Hence, you need to provide it as a column in the CSV file, as shown below:
[image]

To use a custom record ID, map the first field in Column mappings section with the appropriate record ID in your CSV file as shown below. 

[image]

If your Quick Base table uses a custom record ID, whether the CSV file creates or updates the table depends on the following criteria: 

 - If record ID is present in the row, search for the matching record ID in Quick Base then `update` that record. If no matching record ID is found in Quick Base, `create` a new Quick Base record with that record ID.
 - If record ID is empty, that CSV row will `fail`.

PR #633, opened 2018-10-16
Updated images for account property

marygabriola:
1) updated account property navigation gif
2) updated account property data tree png
3) updated create property gif
4) updated update/delete property gif


graceliching:
Hi @marygabriola the updated create property gif does not show the creation step. Can you update please?

PR #644, opened 2018-10-24
Recipe Development Cycle Updated

marygabriola:


marygabriola:
@graceliching Sure. I have added content for truncate table action. I also noticed that there is no content for update and delete do you want me to add these too?

graceliching:
@marygabriola yes, thanks :+1

PR #645, opened 2018-10-24
Fix Summary to point to Google Sheets v3 docs

zweedao:
Temporary, until we release the new v4 actions.


zweedao:
@graceliching you are right. I changed the URL for backward compatibility.

PR #653, opened 2018-11-19
Recipe Development Lifecycle

marygabriola:
Updated screenshots and misspelled words


graceliching:
Hi, other than the mispelled words and the navigation screenshot, everything else in the images are the same. Is there a reason you changed all of them?

marygabriola:
Hi, since I changed these two mandatory updated images:
1) navigate-to-recipelifecycle.gif - (changed it since the navigation menu tab is updated. etc, View all tools from - all tools)
2) import-packages-select-folder2.gif - (changed it to a gif instead of a png to show easy navigation)

I thought it would be better to update all images besides these since the account used was mine for these two. Hehe,

PR #655, opened 2018-11-19
Updated Message Template

marygabriola:
Updated images in the message template Article


graceliching:
@marygabriola this PR is good, but in future, if there are very minor changes (the message template images), you can move on to another article to be more efficient!

PR #684, opened 2018-12-06
File streaming docs

eeshansim:


features/file-streaming.md
```diff
@@ -1,23 +1,92 @@
...
 
 The following diagram illustrates how this works:
 
-![File streaming](/assets/images/features/file-streaming/streaming-graph.png)
+![File streaming](/assets/images/features/file-streaming/streaming-illustration.png)
 
-## 2. How to use it
-Streaming is enabled by default in several Workato file connectors (e.g. Amazon S3, SFTP, On-prem file, etc), and requires no setup.
+## How to use it
+File streaming is an automatic feature. It is initiated when you map a <kbd>File contents</kbd> datapill to a **File contents** input field that both supports streaming.
```
jandonyada: I think we should be upfront about the fact that streaming only works when both the source datapill and the destination input field support streaming. Right now, it's sort of hidden at the end of the 2nd sentence.

"File streaming is enabled by default. It is initiated when you map a <kbd>File contents</kbd> datapill (from the source app) to a **File contents** input field (of the destination app)."

Went out on a limb here, but drawing from the Source <> Workato <> Destination image, I've added some additional stuff in parentheses for greater clarity, but lemme know what you think.

features/file-streaming.md
```diff
@@ -1,23 +1,92 @@
...
 
-Whenever you transfer a file from one system to another, Workato will use streaming if possible. Note that currently, there is a [job](https://docs.workato.com/recipes/jobs.html) runtime limit of 2000 seconds, so the maximum file size you can transfer is about 300MB.
+If either one of the datapill and input field is not streaming-enabled, a standard data mapping is performed - the entire file is downloaded and uploaded as a single string. In this case, the size limit for the action is limited by the respective application.
```
jandonyada: Some of the terms I think can be made a bit more user-friendly:

- streaming-enabled
- as a single string

Once again, drawing from the image:

"If either the source datapill or the destination input field does not support streaming, then standard data mapping is performed, i.e. the entire file is downloaded and then uploaded without any chunking. In this case, the file transfer is limited the size and memory limits imposed by the respective applications(?)."

Also, are the size/memory limits imposed by both the source and destination apps, or just 1 of them? Wasn't sure about this part.

features/file-streaming.md
```diff
@@ -1,23 +1,92 @@
...
 # File Streaming
-Enterprises often have the need to transfer large files of a few hundred megabytes from one system to another. For example, copying daily database dumps from Amazon S3 to SFTP file storage.
+File streaming is the concept of reading and writing a file in smaller parts (chunks) in a sequence. This allows us to transfer large files between applications or file systems without worrying about hitting size or memory limits of the sending and receiving system. A common example is transferring records from a shared file system (SFTP) to a file hosting platform to for analysis (Amazon S3).
 
-Traditional approaches to file transfer usually have a small file size limit or require large server for processing. With Workato, the `File streaming` mechanism allows you to transfer large files between different systems easily, without any setup.
+## How it works
+When transferring contents of a file (download then upload), Workato automatically splits the file into smaller chunks. These chunks are sent to the destination application in separate requests. This approach allows us to stay within the size or memory limit of the application, regardless of the total size of the file.
```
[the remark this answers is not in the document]
eeshansim: Sounds good

features/file-streaming.md
eeshansim: It is at the action level

features/file-streaming.md
eeshansim: Smaller of the 2 size limits

features/file-streaming.md
eeshansim: Phrasing the condition as the first part of the first sentence sounds a little weird. I've rephrased to this:

> File streaming is supported only if both the source download and destination upload actions support this feature. It is automatically initiated when you map a <kbd>File contents</kbd> datapill (from the source app) to a **File contents** input field (of the destination app).

PR #690, opened 2018-12-12
add SAP supported versions

zweedao:


zweedao:
@graceliching one more SAP update, can you help to review :)

graceliching:
@zweedao I've edited it abit. The link you provided is only accessible upon login to SAP, is there one that is accessible publicly with the same info?

zweedao:
we haven't found. That link is supposed to be for SAP customers, so they need to login to view.

PR #701, opened 2018-12-20
update setting for SAP Idoc

zweedao:


zweedao:
@graceliching very quick change, after code review.

PR #702, opened 2018-12-20
Create fixed-dynamic lists.md

graceliching:


features/fixed-dynamic lists.md
```diff
@@ -0,0 +1,48 @@
...
+date: 2018-12-06 16:00:00 Z
+---
+
+# Fixed and dynamic lists
+
+## Using list data pills in actions
+
+Some actions will accept a list as input e.g. Salesforce bulk insert and bulk update actions, create journal entry actions for accounting apps, create sales order actions for ERP apps.
+
+Actions that take lists as inputs will have a input field called Source list, which will only take in list pills. List data pills are indicated with a stack of lines as an icon. When these list input fields are selected, the data tree will offer only list pills.
```
zweedao: 2 things:
- Actually some list-input action doesn't show `Source list` and Dynamic list by defaut. That's why users don't know they can use Dynamic list to handle list output. They just map fields directly to Fixed list and don't know why they only get the first record.

- I think we should point out the list datapill & list input field in the screenshot below. Screenshot has too many elements, not sure where to focus on.

features/fixed-dynamic lists.md
```diff
@@ -0,0 +1,48 @@
...
+*Rows list object is mapped into the Salesforce Contacts input source list*
+
+When actions offer a list as input, the input modes allowed are `Fixed list` and `Dynamic list`. Select the input mode by clicking on the ellipses icon as shown below.
+
+![Dynamic or fixed](/assets/images/features/list-management/dynamiclist.png)
+*Select the input mode*
+
+## Dynamic lists
+
+'Dynamic list' is the default option where a list input is required. Use a dynamic list if the number of items or their content is dependent on another list (for example, the items in an invoice and their prices).
```
zweedao: Same here, Dynamic List is not often the default option.

I think our definition of "if the number of items or their content is dependent on another list" is quite difficult to imagine. The example doesn't make it easier either. 

I think the basic difference between Dynamic List vs Fixed List to user is:
- Dynamic List: help you automatically loop through a list of records (such as list of CSV rows, list of records from Search action)
- Fixed List: you need to add each record one by one

features/fixed-dynamic lists.md
```diff
@@ -0,0 +1,48 @@
...
+
+When this option is selected, the first field to map is the 'Source list' field, which only accepts a list data pill. Using the list data pill will read directly from the source application CSV file.
```
zweedao: CSV file was not mentioned anywhere before this sentence. Not clear here.

features/fixed-dynamic lists.md
```diff
@@ -0,0 +1,48 @@
...
+
+Thereafter, map the fields below using the pills nested within the list pill. Use the 'Dynamic list' option when you want the action to iterate through the list line by line, using the values in each row in all the fields you map here.
+
+In the example below, the source list comes from a CSV file uploaded into Dropbox and will create Salesforce Opportunity records in bulk using the data in each line of the CSV file. Thus, each line in the CSV file will provide data per Opportunity record created. Each Salesforce field is mapped to a column in the CSV file.
```
zweedao: I think we should explain how it works more clearly:
1. You provide the list data pill `CSV rows` to the `Source list` input field
2. Dynamic List will then automatically loop through all those `CSV rows` and read the column values from each row, such as Country, Name, Age, etc. Then input to Salesforce.

features/fixed-dynamic lists.md
```diff
@@ -0,0 +1,48 @@
...
+
+![Example input source list](/assets/images/features/list-management/example-input-source-list.png)
+*Rows list object is mapped into the Salesforce products input source list*
+
+
+## Fixed lists
+
+In contrast, using the fixed, or static list input requires either text input from the user or data pills from a source that is not a list.
```
zweedao: Same like above. This definition is not very either to imagine. 

Fixed List basically means you need to add record one by one instead of automatic looping.

features/fixed-dynamic lists.md
```diff
@@ -0,0 +1,46 @@
...
+
+In the example below, the source list comes from a CSV file uploaded into Dropbox and will create Salesforce Opportunity records in bulk using the data in each line of the CSV file. It will loop through all CSV rows and read the column values from each row, such as Product code, Product description etc. Thus, each line in the CSV file will provide data per Opportunity record created in Salesforce. 
+
+![Example input source list](/assets/images/features/list-management/example-input-source-list.png)
+*Rows list object is mapped into the Salesforce products input source list*
+
+
+## Fixed lists
+
+In contrast, using the fixed, or static list input requires either text input from the user or data pills from a source that is not a list. Instead of using a list and looping through it automatically, use a fixed list to add records one by one.
```
zweedao: @graceliching How about this:

In contrast, use the 'Fixed list' option when you want to input a list with a fixed number of items, without dynamically iterating through another list. In this case, you need to add each item one by one.

features/fixed-dynamic lists.md
```diff
@@ -0,0 +1,46 @@
...
+
+
+## Fixed lists
+
+In contrast, using the fixed, or static list input requires either text input from the user or data pills from a source that is not a list. Instead of using a list and looping through it automatically, use a fixed list to add records one by one. 
+
+![Dynamic or fixed](/assets/images/features/list-management/fixedlist.png)
+*Select the input mode*
+
+Use this input mode if the total number of items in the list and their contents should be the same across recipe jobs.
```
zweedao: No need this sentence. My suggested sentence above already covered this.

graceliching:
@zweedao I've edited based on your comments

features/fixed-dynamic lists.md
```diff
@@ -0,0 +1,46 @@
...
+
+The first example below uses text input to enter the First Name and Last Name of a Salesforce contact. In this case, the user knows the names and the number of contacts they wish to create in Salesforce.
```
zweedao: The example below demonstrates using Fixed list to create a fixed number of Contacts in Salesforce. In this case, the user already knows the number of contacts he wishes to create in Salesforce, so he will add each item in the `Contacts` list one by one.

zweedao:
@graceliching some more comments :)

graceliching:
@zweedao Thanks, I've edited

PR #704, opened 2018-12-21
Put SAP docs into SUMMARY

zweedao:


zweedao:
We finally released SAP to production. Putting doc in Summary @graceliching :)

PR #706, opened 2019-01-04
New SF Batch actions

graceliching:


connectors/salesforce/batch-operations.md
```diff
@@ -0,0 +1,122 @@
...
+
+![Select object](/assets/images/salesforce-docs/salesforce-batch-select-object.gif)
+*Select Salesforce Contact object*
+
+The list input pill can come from various sources such as a new CSV file in a file storage app (eg. Box, Amazon S3, On-prem files) or a list created with the Lists by Workato application. To learn more on how to use list inputs, refer to the [List Management](/features/list-management.md) documentation.
+
+![Select object](/assets/images/salesforce-docs/salesforce-batch-select-object.png)
+*Select list input pill*
+
+Thereafter, proceed to select the rest of the optional Salesforce fields that need to be mapped with data in the dropdown list at the end of the action and map the fields.
```
zweedao: Thereafter, in the `Add/remove optional fields` section at the end of the action, proceed to select any additional optional Salesforce fields that you need then map those fields.

connectors/salesforce/batch-operations.md
```diff
@@ -0,0 +1,122 @@
...
+
+![Select optional fields](/assets/images/salesforce-docs/batch-select-optional-fields.png)
+*Select optional fields*
+
+### Advanced configuration
+
+![Advanced configuration](/assets/images/salesforce-docs/batch-advanced-config.png)
+*Advanced configuration*
+
+In the advanced configuration section, choose if the request should be rolled back if a record fails to be updated. The **entire** request will be rolled back if the list has less than 200 lines (less than 200 records created in Salesforce). This means that if 1 of 200 records fail, all the records will not be updated.
```
zweedao: In the advanced configuration section, choose if you want to roll back if the record creation fails. If **Yes**, this is how the roll-back works:

Workato automatically divides the record creation into smaller batches of 200 records/batch. So when 1 record in a batch fails, that whole batch will be rolled back, but other batches will not be affected.

For example, if you are creating 1000 records, they will be divided into 5 batches of 200 records each. If a record in the 1st batch failed, that whole 1st batch of 200 records will be rolled back. However the other 4 batches will not be affected and will be processed as normal.

connectors/salesforce/batch-operations.md
```diff
@@ -0,0 +1,122 @@
...
+
+However, if the batch is larger than 200 records, only the affected batch is rolled back. For example, if the list is 1,000 lines long, it is processed in 5 batches of 200. If the 1st line fails, only the 1st batch out of 5 is rolled back.
```
zweedao: replaced by above comment

graceliching:
@zweedao thanks, I have addressed the comments. Could you approve?

PR #714, opened 2019-01-18
Prepare for SAP RFC release

zweedao:


zweedao:
quick fix @graceliching :)

PR #727, opened 2019-02-13
Change SAP default parameter values

zweedao:


zweedao:
quick one @graceliching , can you help review & merge :)

PR #728, opened 2019-02-13
Exponential operator

eeshansim:


formulas/number-formulas.md
```diff
@@ -72,6 +72,22 @@ Divides left hand operand by right hand operand.
 
 ---
 
+## The exponential (\*\*) operator
+
+Divides left hand operand by right hand operand.
```
graceliching: hahaha i think this should be: 
Right-hand operand is the exponent of the left-hand operand.

PR #752, opened 2019-03-20
Snowflake database timezone and trigger

eeshansim:


bennettgo:
I made some very minor grammar changes as i was reading the 2 pages! Other than that it seems clear to me!

eeshansim:
> Looks good! though any reason why the SQL commands were removed?

Does not apply to Snowflake

PR #755, opened 2019-03-22
Added Workbot for MS Teams to root

jandonyada:


SUMMARY.md
```diff
@@ -151,6 +151,7 @@
     * [Configuring slash commands](workbot/configuring-slash-commands.md)
     * [Legacy slash commands](workbot/legacy-slash-commands.md)
   * [Workbot for MS Teams](/workbot-for-teams/workbot.md)
+    * [Getting started](/workbot-for-teams/getting-started.md)
```
eeshansim: We shouldn’t put the same doc in 2 places in summary.md, it messes up the nav section

jandonyada: I will remove it from Tools

PR #772, opened 2019-04-11
SAP: update library instruction

zweedao:
SAP: update library instruction
fix https://github.com/workato/sap_java/issues/72


zweedao:
@graceliching can you help review & merge

PR #792, opened 2019-05-14
Flowchart fix

bennettgo:


eeshansim:
The image looks a little blur and this part (circled in red) is cropped out

flowchart

PR #813, opened 2019-05-27
SAP: fix on gwser instruction

zweedao:


zweedao:
Quick one @graceliching , can you help review & merge 👌

zweedao:
@graceliching so far we see both forms (text vs number) in different customers. The way we build the connector is not very definite yet. So need to put this in the doc as a quick fix for now.

PR #837, opened 2019-06-25
Update workbot-triggers.md

ngchinhow:
Moved the section about advanced parameters passing to the new passing parameters article.

### DIFF
--- a/workbot-for-teams/workbot-triggers.md
+++ b/workbot-for-teams/workbot-triggers.md
@@ -1,6 +1,6 @@
 ---
 title: Workbot for Teams triggers
-date: 2017-03-30 05:00:00 Z
+date: 2019-06-24 05:00:00 Z
 ---
 
 # Workbot command trigger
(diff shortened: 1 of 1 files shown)


workbot-for-teams/workbot-triggers.md
```diff
@@ -70,36 +70,24 @@ Workbot commands can invoke their recipes by:
 *The 'Create issue' button invokes the 'newissue' command and executes the recipe when submitted*
 
 #### Parameters
-Each parameter can store additional data that can be used as datapills in follow-up recipe actions.
+The user or another recipe can provide additional information to a recipe by [passing parameters](/workbot-for-teams/passing-parameters.md) to that recipe. This requires the parameters to be defined in the Workbot trigger, after which the corresponding datapills will be available for use in the follow-up recipe actions. It must be noted that without defining these parameters, their corresponding datapills will not be available for the recipe actions, even if a valid value was passed to it.
```
jandonyada: The description is generally good, but I think it can be improved further.

>The user (or a preceding recipe) can provide additional information to the recipe by passing parameters to the recipe.
>
>This requires the parameters to be defined in the Workbot trigger, after which the corresponding parameter datapills will be available for use in subsequent recipe actions.
>
>Its important to note that without defining parameters, the parameter datapills will not be available for subsequent recipe actions — even if valid values were passed to it.

jandonyada: What do you think?

ngchinhow: The user (or a preceding recipe) can provide additional information to a recipe by passing parameters to it.

This requires the parameters to be defined in the Workbot trigger, after which the corresponding parameter datapills will be available for use in subsequent recipe actions. If this step is not done, no parameter datapills will appear for use. All parameter values passed to this recipe, however valid, will not be captured, since there is no place for it to go.

ngchinhow: I was afraid that my original phrasing made it sound like I was just saying the same thing twice, just from opposite directions. Hope I changed it to be clearer.

jandonyada: That's fine. Improving on your suggestion:

First part is good to go.
>The user (or a preceding recipe) can provide additional information to a recipe by passing parameters to it.

Made some edits to the 2nd part. I'm trying to shorter, more structured paragraphs, while adding in the fact that the parameter names should be the same when passing parameters from an upstream recipe to a downstream recipe:
>This requires the parameters to be defined in the Workbot trigger, after which the corresponding parameter datapills will be available for use in subsequent recipe actions. This step is critical — it ensures that:
>
>1. Parameter values are passed to the corresponding parameters. If you're passing parameters from a preceding recipe, make sure the parameter in the upstream recipe and the parameter in the downsteam recipe have same name. This ensures that the correct parameter value is passed to the correct parameter.

>2. Parameter datapills become available in the datatree for use in subsequent recipe actions.

ngchinhow: That feels like the steps are just repeating the second paragraph. Maybe we can try combining them together.

ngchinhow: This requires defining the parameters (link to defining parameters) in the Workbot trigger, ensuring that the parameter's name is the same in both upstream and downstream recipes. 

// A nuance is that the parameter name in the upstream recipe (in additional parameters and choice parameter) must be all lowercase. The parameter name defined in the downstream recipe does not have any restrictions to how its name is spelt, so long as the alphabets are the same. This is a really weird nuance to have and can make this parameter passing thing more confusing for users.

This step is critical - it makes sure that:

1. The correct parameter value to passed to the correct parameter through the recipes. 

2. Parameter datapills become available in the downstream recipe's datatree for use in subsequent recipe actions.

jandonyada: The uppercase/lower case thing is a bug - it should receive any case and be converted to lowercase. Ok let's go with your last edit.

jandonyada: Commit this and we can publish

workbot-for-teams/workbot-triggers.md
```diff
@@ -70,36 +70,29 @@ Workbot commands can invoke their recipes by:
 *The 'Create issue' button invokes the 'newissue' command and executes the recipe when submitted*
 
 #### Parameters
-Each parameter can store additional data that can be used as datapills in follow-up recipe actions.
+The user (or a preceding recipe) can provide additional information to a recipe by passing parameters to it.
 
-For example, to create an incident in ServiceNow, you may want to prompt users for additional info like **Urgency**, **Summary** and **Description**. By adding **Urgency**, **Summary** and **Description** as parameters, Workbot will open a task module and prompt the user for each parameter.
+This requires defining the parameters (link to defining parameters) in the Workbot trigger, ensuring that the parameter's name is the same in both upstream and downstream recipes.
```
ngchinhow: I think we should add the recipe-chaining.png image from passing-parameters.md here as well.

jandonyada: Added.

PR #850, opened 2019-07-02
Adding docs guide to repo

graceliching:
For easy navigation to guide


eeshansim:
@graceliching I feel that this is more appropriate in Wiki instead of being part of the repo content

graceliching:
yep that works, but we have to make it known that it exists there. I don't think people usually navigate to the wiki unless they know content is available

PR #853, opened 2019-07-10
Update sample bash commands

eeshansim:


ananich:
Usually commands that are executed by non-privileged users start from `$`
And commands that requires root privileges starts with `#`

PR #883, opened 2019-08-09
Update Collections

gabrielsim:
Removing `RIGHT JOIN` as they are not currently supported.

### Which issues does this fix?
Fixes #12345

### PR submission checklist
- [ ] Optimised every new image
- [ ] Checked all content for grammatical errors (using Grammarly, for example)
- [ ] Added new files to SUMMARY.md

PR #884, opened 2019-08-09
removed-right-join

eeliang:
Removed `right-join`


eeliang:
https://github.com/workato/docs/pull/883 addresses the same issue.

PR #900, opened 2019-09-13
Fixed the API doc

kandadaboggu:


eeliang:
New changes:
- Applying `isTocVisicle: true` to all pages
- Updating endpoints to use `blockquotes`
- Differentiating **URL parameters** and **Request body**
- Removing error code from individual pages, error codes will be shown in workato-api.md

PR #903, opened 2019-09-13
Added styles for API input and API quick reference tables

kandadaboggu:
Added support for fixed col width for tables for displaying API parameters.

The markdown engine used by Gitbook doesn't support mixed mode
```
<div class='api_input'>
| Type  | Resource  | Description                   |
|-------|-----------|-------------------------------|
| POST  | `users`   | Create a new OEM user.        |
| POST  | `account` | Add member to the OEM account.|
</div>
```

There are new markdown parsers like https://github.com/gettalong/kramdown supports ability `markdown="1"` . Gitbook doesn't support this feature.

This approach side loads an invisible `div` when we want to target a container for styling. The styling is done using immediate sibling selector in CSS.

### markdown

```
<div class='api_input'></div>

| Type  | Resource  | Description                   |
|-------|-----------|-------------------------------|
| POST  | `users`   | Create a new OEM user.        |
| POST  | `account` | Add member to the OEM account.|

```

### css

```CSS
div.api_input + table {
  width: 100%;
}
```


styles/website.css
```diff
@@ -38,6 +38,57 @@ body {
...
 }
 
+/* In page TOC */
+.intopic-toc h3 {
+  color: #333;
+  font-size: 1.1em;
+}
+
+/* Parameter table */
+div.api_input + table {
```
th0r: Let's just leave a class here and all other such places: `.api_input + table`

styles/website.css
```diff
@@ -38,6 +38,57 @@ body {
...
+  width: 100%;
+}
+
+/* Name */
+div.api_input + table tr th:nth-child(1) {
```
th0r: Do you need `tr` here? `.api_input + table th:nth-child(1)` should be enough.

styles/website.css
kandadaboggu: What do you mean? Do you mean we can do this?

```css
div.api_input {
  width: 100%;
}
```

kostos:
```suggestion
.api_input + table {
```

styles/website.css
```diff
@@ -38,6 +38,57 @@ body {
...
+  font-size: 1.1em;
+}
+
+/* Parameter table */
+.api_input + table {
+  width: 100%;
+}
+
+/* Name */
+.api_input + table tr th:nth-child(1) {
```
kostos: Remove `tr`. Check code below, it also have same issues
```suggestion
.api_input + table th:nth-child(1) {
```

styles/website.css
```diff
@@ -38,6 +38,57 @@ body {
...
+  width: 60%;
+}
+
+.api_quick_reference + table {
+  width: 100%;
+}
+
+/* Type */
+.api_quick_reference + table tr th:nth-child(1) {
+  text-align: left;
```
kostos: All rules repeating `text-align: left;`

you can make 
```css 
.api_input th, .api_quick_reference {
  text-align: left;
}
```

And specify width only

kandadaboggu: table is sibling to `.api_input`

PR #908, opened 2019-09-19
Block kit docs

jandonyada:


workbot/block-kit.md
```diff
@@ -0,0 +1,90 @@
...
+
+[Block Kit](https://api.slack.com/block-kit) is Slack's UI framework that allows more control and flexibility when building messages for Slack.
+
+The [Post command reply](/workbot/workbot-actions.md#post-command-reply) and [Post message](/workbot/workbot-actions.md#post-message) actions both support block kit. You can stack "blocks" and customize the order and appearance of each block, as well as the elements within each block.
+
+You can also optionally provide secondary attachments, which will display below any defined blocks in the Post command reply or Post message actions.
+
+# Block kit previewer
+Preview the messages you've built with the Post message or Post command reply actions by clicking on **See preview on block kit builder** in the block hint.
+
```
bennettgo: Should there be a small paragraph to teach users how to re-order blocks or if it is possible to dynamically generate blocks instead of a static list? 
I was just thinking it might be something to add specially since I'm assuming order of fields in block kit matter quite abit

jandonyada: Re-ordering a static list of blocks can be done by drag-dropping. You can dynamically generate blocks too.

So yea I think a paragraph about static / dynamic blocks is a good idea. I'll add it in soon.

PR #931, opened 2019-10-15
HOWTO encrypt connection between OPA and Microsoft SQL Server

ananich:
Published option that allows to encrypt connection between OPA and Microsoft SQL Server

### DIFF
--- a/on-prem/agents/profile.md
+++ b/on-prem/agents/profile.md
@@ -91,7 +91,7 @@ A database type is specified either by using the `adapter` property or a complet
 
 `port` numbers can be omitted when matching defaults for a given database type.
 
-Here's a sample SQL server configuration for connecting to a specific instance:
+Here's a sample Microsoft SQL Server configuration for connecting to a specific instance:
 
 ```YAML
 database:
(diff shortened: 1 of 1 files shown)


eeshansim:
@ananich what does `encrypt` property default to?

ananich:
Default to use unencrypted connection.

on-prem/agents/profile.md
```diff
@@ -102,6 +102,7 @@ database:
     database: sales
     username: me
     password: foobar
+    encrypt: false
```
eeshansim: > Default to use unencrypted connection.

In that case, should the example be `encrypt: true` then?

ananich: In case we use encryption by default, the SQL Server instance must be configured to encrypt. This is a prerequisite. There is a bunch of potential issues here:
1) user will need a valid SSL certificate
2) should reference the server by IP or DNC name which matches one pinted into the certificate
3) in case they use a custom Certificate Authority (most of corporations do so), they need to add Root CA certificate to JVM's trust store.

PR #936, opened 2019-10-21
Added support for shareable pages. Added embedding connections doc.

n0mid:
Added documentation for Embedding Connections 

Documentation link is `oem/embedding-connections.html`.

Снимок экрана 2019-10-21 в 13 22 19


oem/embedding-connections.md
```diff
@@ -0,0 +1,59 @@
...
+
+    function receiveMessage(event) {
+       var data = JSON.parse(event.data);
+       
+       switch (data.type) {
+           case 'heightChange':
+              document.getElementById('workatoId').style.height = data.payload.height + 'px';
+              break;
+           case 'connectionStatusChange':
+               var message = data.payload.connected ? 'Connected' : 'Disconnected';
```
kandadaboggu: Fix the alignment of the code.

oem/embedding-connections.md
```diff
@@ -0,0 +1,59 @@
...
+               break;
+   
+       }
+    }
+  </script>
+</head>
+
+<body>
+<h4>Status: <span id="statusId"></span></h4>
+<iframe id="workatoId" src="https://workato.com/direct_link/embedding_connections/<connection_id>>?workato_dl_token=<jwt_token>" style="width: 500px; height: 150px; border: 0"></iframe>
```
kandadaboggu: Can we examples of how dl_token can be generated?

oem/embedding-connections.md
```diff
@@ -0,0 +1,82 @@
...
+Widget API works on postMessage. 
+
+Format: `{type: string, payload: object}` 
+
+Is supported next types: 
+
+| Type  | Payload | Description |
+| ------------- | ------------- | -------- |
+| heightChange  | { height: number } | Content height was changed
+| connectionStatusChange  | { connected: string, connectionId: number, provider: string } | Connection status was changed
```
kandadaboggu: - We should return the error message if any.
- What are the possible values for connected? 
- Can we make the payload similar to the payload we return in our public API's

```ruby
{
  id: 123,
  provider: "salesforce",  
  authorization_status: "failure", #success/failure
  authorization_error: "Unauthorized"
}
```
https://docs.workato.com/workato-api/managed-users.html#list-customer-connections

n0mid: My mistake, `connected` is a boolean. 

Something like this? 
```
{
  id: <connection_id>,
  provider: <provider>,  
  connection_status: <connected/disconnected>
  connection_error: <err_message/null>
}
```

kandadaboggu: okay. Looks fine

kandadaboggu: You can use `connected` (true/false), do add a field for `error`.

PR #937, opened 2019-10-22
Update role privilege table

graceliching:
Fixes https://github.com/workato/issues/issues/10097


graceliching:
There's a merge cell in the middle of the table and the colspan was useful in formatting the width of the table. I changed it to mostly be a markdown table, but the last table at the end 'folder access' still has to be formatted with HTML otherwise it'll be in bold

PR #959, opened 2019-11-25
Increase search index size

### LINKED ISSUE #958: Search is broken
Quote from Slack:

Not sure if this is the right channel but the search on docs.workato.com seems broken (or just un-intuitive). E.g. wanting to know what our IP addresses to whitelist: https://docs.workato.com/?q=whitelist doesn't return the obvious right result: https://docs.workato.com/security/ip-whitelist.html (is it because it's a sub-section?). Had to google to find the right page every time.

th0r:
### Which issues does this fix?
Fixes #958 

### Before:
image

### After:
image


th0r:
Irrelevant after #968

PR #960, opened 2019-11-26
Add section for general traffic TO Workato and IP address difference

devenmaru:
DOcumentation update based on suggestion from Dimitris to add section on general traffic (like user browser) to Workato and how the IP addresses of those servers may be different from the traffic that's generated FROM Workato like on prem agent.


devenmaru:
@dkogias we can remove reference to support if there's no role for support to play here.

PR #979, opened 2019-12-16
Fix issues with Algolia search bar

th0r:
1. Hides original search bar if Algolia search is enabled
2. Fixes styles for Algolia search bar

image

### DIFF
--- a/docs/.vuepress/theme/components/Search.vue
+++ b/docs/.vuepress/theme/components/Search.vue
@@ -1,8 +1,8 @@
 <template>
   <div class="Search">
-    <i class="fa fa-search" aria-hidden="true"></i>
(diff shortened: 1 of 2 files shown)


docs/.vuepress/theme/components/Search.vue
```diff
@@ -1,8 +1,8 @@
 <template>
   <div class="Search">
-    <i class="fa fa-search" aria-hidden="true"></i>
+    <i class="fa fa-search" aria-hidden="true"/>
```
n0mid: Is empty tag valid for `i`? Or vueJs will convert it to normal html?

th0r: Yep, it will convert it.

th0r:
@eeshansim can we merge this please?

Issue #984, opened 2019-12-17
Too long delay before the page scrolls back to the top after change


demeshkin:
1. Go to https://docs.workato.com/security/two-factor-authentication.html
2. Scroll all the way to the bottom of the page
3. Select `IP whitelist` from the menu. You'll see the bottom of the page and after several seconds the page scrolls to the top.

![2019-12-17 09 31 33](https://user-images.githubusercontent.com/522683/70957128-12cd5180-20b0-11ea-805d-c2e5b9cebd0a.gif)

th0r:
That's how VuePress works. If you have any ideas how to fix it feel free to create a PR.

demeshkin:
@th0r What do you think about that? https://github.com/workato/docs/pull/988

demeshkin:
@th0r First navigation is also quite weird.

![2019-12-18 12 01 35](https://user-images.githubusercontent.com/522683/71054854-6dd57600-218e-11ea-8cd0-5f76c739cad5.gif)

PR #988, opened 2019-12-18
Removed smooth scrolling

demeshkin:
I think it makes navigation much faster.

closes https://github.com/workato/docs/issues/984

### DIFF
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -31,7 +31,7 @@ module.exports = {
   ],
   themeConfig: {
     sidebar: require('./sidebar'),
-    smoothScroll: true,
+    smoothScroll: false,
     searchPlaceholder: 'Type to search Workato docs',
     searchMaxSuggestions: 10,
     // Is used to generate URL for "Edit this page" link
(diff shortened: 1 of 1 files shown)


demeshkin:
@th0r But now scroll to the particular section is broken.

demeshkin:
@th0r Actually it doesn't work with smooth scrolling too. Probably some bug of `vuepress` in dev mode.

th0r:
So what do you want me to do?

demeshkin:
@th0r Can you review it?

th0r:
> But now scroll to the particular section is broken.

What about this?

demeshkin:
@th0r See https://github.com/workato/docs/pull/988#issuecomment-566870232

th0r:
> Actually it doesn't work with smooth scrolling too

What do you mean? Can you reproduce it?

demeshkin:
@th0r Yes. When I run it in dev environment I see this.

docs/.vuepress/config.js
```diff
@@ -31,7 +31,7 @@ module.exports = {
   ],
   themeConfig: {
     sidebar: require('./sidebar'),
-    smoothScroll: true,
+    smoothScroll: false,
```
th0r: Let's add comment about why we decided to no use smooth scroll (delay before scroll to top + slow scrolling on pages with a lot of content).

demeshkin: Added

Issue #986, opened 2019-12-17
strftime docs have characters that render as emojis


wchijie:
In the strftime formula section of our docs, some characters are rendering as emojis.

![image](https://user-images.githubusercontent.com/15618016/71022380-4ef9c400-213b-11ea-98e5-bb61425b6c20.png)

https://docs.workato.com/formulas/date-formulas.html#strftime

PR #989, opened 2019-12-18
Disable Emoji markdown plugin

th0r:
Fixes https://github.com/workato/docs/issues/986

image


docs/.vuepress/config.js
```diff
@@ -44,6 +44,10 @@ module.exports = {
   markdown: {
     // Searches for URLs in plain text and converts them to links
     linkify: true,
+    chainMarkdown: config => {
+      // Fixes https://github.com/workato/docs/issues/986
+      config.plugins.delete('emoji');
```
kandadaboggu: Does this mean we can't use emoji?

th0r: Well, yes. Do we want to?

th0r: If we want then I think there should be a way to escape them.

th0r: @kandadaboggu seems like they don't support escaping: https://github.com/markdown-it/markdown-it-emoji/issues/11

So we'll need to wrap all false-positives with backticks (code).

kandadaboggu: We should escape them.

th0r: I've found a way to disable emoji aliases like `:)` but still enable full-sized emojis like `:smile:`

kandadaboggu: Okay

PR #996, opened 2019-12-27
Coupa schema update

bennettgo:


bennettgo:
@eeshansim Compressed the images!

PR #1012, opened 2020-01-20
added binary file download example

bennettgo:

### DIFF
--- a/docs/developing-connectors/sdk/action.md
+++ b/docs/developing-connectors/sdk/action.md
@@ -176,10 +176,21 @@ Each action should have a `Test` button which would allow you to manually declar
 > Console logs come in useful with ruby <code>puts</code> functions. You can use this to debug your actions and triggers easily
 
 ## Variations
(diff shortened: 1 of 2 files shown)


docs/developing-connectors/sdk/action.md
```diff
@@ -211,9 +222,37 @@ execute: lambda do |connection, input|
     )
 end
 ```
-
 See [Methods](/developing-connectors/sdk/methods.md) section for list of methods available for use in your custom connector actions.
 
+### Configuring actions which download binary files
+As Workato expects data in JSON formats unless told otherwise, download actions which work with Binary files require you to tell Workato to expect the response in a raw format.
+
+Below we have an example from the Egnyte API, where we want to download files of any content-type.
```
eeshansim: The example has a lot of other lines. Can we draw more attention to `response_format_raw` here?

bennettgo: removed peripheral methods!

PR #1024, opened 2020-02-02
API table styles

graceliching:
@th0r Hi Yuriy, can you help take a look at this PR and see why the table styles do not apply to some tables:

Eg. below where table width should be 100%: 
image

Thank you!

### DIFF
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -1,6 +1,9 @@
 const {resolve} = require('path');
 const {repository, domain} = require('../../package.json');
 
+const attributesPlugin = require('markdown-it-attrs');
+const imageSizePlugin = require('./plugins/markdown-it-image-size');
+
 const IMAGES_ALIAS = '@img';
 const IMAGES_DIR = resolve(__dirname, '../../assets/images');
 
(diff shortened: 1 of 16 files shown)


graceliching:
I think we should disable the auto-scaling of the table to a shorter width, it makes multiple tables in the same article look messy
image

th0r:
@graceliching fixed

th0r:
@graceliching what do you think about this plugin? https://github.com/arve0/markdown-it-attrs
It allows to set any attributes to any markdown tag. In case of tables it will look like this:
```md
| Name  | Description  | Sample reply                  |
|-------|--------------|-------------------------------|
| `200` | Success      | `{"Success": true}`           |
| `400` | Bad request  | `{"message": "Bad request"}`  |
| `401` | Unauthorized | `{"message": "Unauthorized"}` |
| `404` | Not found    | `{"message": "Not found"}`    |
| `500` | Server error | `{"message":"Server error","id":"3188c2d0-29a4-4080-908e-582e7ed82580"}` |
{.api_input}
```
I tried it and it works.

graceliching:
> @graceliching what do you think about this plugin? https://github.com/arve0/markdown-it-attrs
> It allows to set any attributes to any markdown tag. In case of tables it will look like this:
> 
> ```lisp
> | Name  | Description  | Sample reply                  |
> |-------|--------------|-------------------------------|
> | `200` | Success      | `{"Success": true}`           |
> | `400` | Bad request  | `{"message": "Bad request"}`  |
> | `401` | Unauthorized | `{"message": "Unauthorized"}` |
> | `404` | Not found    | `{"message": "Not found"}`    |
> | `500` | Server error | `{"message":"Server error","id":"3188c2d0-29a4-4080-908e-582e7ed82580"}` |
> {.api_input}
> ```
> 
> I tried it and it works.

Thank you! Great, I think it looks cleaner than the way it is today. @th0r

th0r:
> Thank you! Great, I think it looks cleaner than the way it is today.

Done

PR #1049, opened 2020-02-19
NetSuite get all records action

bennettgo:


docs/connectors/netsuite.md
```diff
@@ -410,6 +410,17 @@ When selecting the fields to search by in the NetSuite connector, take note that
 
 If you can't find the field you're looking for, check with your NetSuite admin or with us!
 
+## Get all standard objects action
+The get all standard objects action will return the entire list of records of the selected record type in your NetSuite.
+
+### Configuring the get all standard objects action
+To carry out a get all standard object action, we would first need to tell the recipe exactly what standard object category you want to retrieve. The output of this action will be a list containing all records in your NetSuite instance. You can then filtering this list of records using `.where` methods in formula mode in downstream actions. [Find out more here](/formulas/array-list-formulas.html#where)
```
graceliching: _To use the 'Get all standard objects' action_, we would first need to tell the recipe exactly what standard object category you want to retrieve. The output of this action will be a list containing all records in your NetSuite instance. You can then _filter_ this list of records using `.where` methods in formula mode in downstream actions. [Find out more here](/formulas/array-list-formulas.html#where).

docs/connectors/netsuite.md
```diff
@@ -410,6 +410,17 @@ When selecting the fields to search by in the NetSuite connector, take note that
...
+
+![Get all standard objects](~@img/connectors/netsuite/get-all-standard-objects.png)
+*Configuring the get all standard objects action*
+
+If you can't find the field you're looking for, check with your NetSuite admin or with us!
```
graceliching: Should this be referring to a standard object instead of a field?

PR #1051, opened 2020-02-20
Update collection.md

ultratin:
remove 50k limit documentation


ultratin:
Based on our increasing ETL usecases and having to recommend collections as the transformation tools. Having that 50k there raise questions about the tool's ability when it actually can load up to millions of rows.

eeshansim:
This doesn't change the limit flag in the product. Please create a github issue.

ultratin:
https://github.com/workato/issues/issues/10685

ultratin:
@eeshansim anything else that needs to be done on my end for this?

PR #1078, opened 2020-03-10
Clarify OPA upgrade step

eeshansim:
Fixes workato/issues#11390

Some users run into an issue because we don't explicitly mention navigating to the agent directory

### DIFF
--- a/docs/on-prem/agents/upgrade.md
+++ b/docs/on-prem/agents/upgrade.md
@@ -16,7 +16,7 @@ Follow the instructions below for upgrading an existing agent:
 
 ## Linux/MacOS
 1. Verify that the agent process is stopped. The upgrade will fail if any running agents are detected.
-2. Navigate to the agent directory. (e.g. `cd ~/programs/workato-agent/`)
+2. Navigate to the agent directory (e.g. `cd ~/programs/workato-agent/`). **Do not** navigate further into the `/bin/` directory.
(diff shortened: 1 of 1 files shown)


docs/on-prem/agents/upgrade.md
```diff
@@ -16,8 +16,9 @@ Follow the instructions below for upgrading an existing agent:
 
 ## Linux/MacOS
 1. Verify that the agent process is stopped. The upgrade will fail if any running agents are detected.
-2. Run the upgrade script: `bin/upgrade.sh`. Make sure you have enough permissions.
-3. Follow the instructions provided by the upgrade script. Confirm the upgrade when prompted.
-4. Upon successful completion of the upgrade, run the agent (e.g. `bin/run.sh`).
-5. Make sure your agent is active and verify its version number on the [on-prem agent page](https://www.workato.com/on_prem_groups).
-6. The upgrade process is not triggered if no new versions are available. However, it might be necessary to repair a broken installation; in that case use the command line option when running the upgrade: `bin/upgrade.sh --enforce`.
+2. Navigate to the agent directory. (e.g. `cd ~/programs/workato-agent/`)
```
wchijie: In this line, we should call out that user should not navigate directly to the bin folder.

Navigate to the agent directory (e.g. `cd ~/programs/workato-agent/`). Do not navigate further into the bin directory.

PR #1093, opened 2020-03-18
SDK - Fix gaps in documentation

bennettgo:


docs/developing-connectors/sdk/object-definition.md
```diff
@@ -473,5 +478,69 @@ object_definitions: {
...
+      "columnName": "column1"
+    },
+    {
+      "columnName": "column1"
+    }
+  ]
+}
+```
+
+This input can be massaged into our desired format with a simple `.map` method.
```
eeshansim: simpler to use `pluck('columnNames')`

docs/developing-connectors/sdk/object-definition.md
```diff
@@ -532,9 +532,7 @@ actions:{
     end,
 
     execute: lambda do |connection, input|
-      column_names = input["columnNames"].map do |item|
-        item['columnName']
-      end || [] # This gives us ["column1","column2","column3"]
+      column_names = input["columnNames"].pluck("columnName") || [] # This gives us ["column1","column2","column3"]
```
eeshansim: It should be this. And there is no need for `|| []` since `pluck` always returns an array.
```ruby
column_names = input.pluck("columnNames")
```

PR #1148, opened 2020-05-19
Edited output of connector API

graceliching:


docs/oem/oem-api/adapters.md
```diff
@@ -30,29 +30,360 @@ curl  -X GET http://www.workato.com/api/integrations \
       -H 'x-user-token: <token>' \
```
eeliang, line 30: Use **https**

